PR #17, opened 2017-09-05
If verification of artifact failed - try to download it again

### LINKED ISSUE #9: If verification of artifact failed - try to download it again

goldmann:
When a file exists on disk and the checksum is not correct
we should try to download that file again and fail
if the checksum after download is still not correct.

Fixes #9

### DIFF
--- a/concreate/tools.py
+++ b/concreate/tools.py
@@ -46,7 +46,7 @@ class Artifact(object):
 
     def _generate_url(self):
         """ Adjust url to use artifact cache if needed.
-        
+
         It can replace:
           #filename# - replaced by a basename of file
           #algorithm# - replace by an algorithm family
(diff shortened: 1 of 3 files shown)


concreate/tools.py
```diff
@@ -65,7 +65,7 @@ def _generate_url(self):
                             .replace('#hash#', self.sums[alg]))
                 break
 
-    def verify(self):
+    def _verify(self):
```
dbecvarik: Even if called during fetch() I would still keep this method "public".

concreate/tools.py
```diff
@@ -92,20 +92,27 @@ def _check_sum(self, algorithm, expected_chksum):
...
+                self._verify()
+                logger.debug("Using fetched artifact '%s' for '%s'. " % (self.filename,
+                                                                         self.name))
+                return self
+            except ConcreateError:
+                # Artifact available locally is not correct.
+                # We need to download it again.
+                pass
+
+        download_file(self.url, self.filename)
```
dbecvarik: We can do this in except block and then have code like:

``` python
             try:
                  self._verify()
                  logger.debug("Using fetched artifact '%s' for '%s'. " % (self.filename,
                                                                           self.name))
              except ConcreateError:
                  # Artifact available locally is not correct.
                  # We need to download it again.
                  download_file(self.url, self.filename)
                  self.verifry()

       return self
```

goldmann: This does not help since we need to download it when the file does not exists too.

concreate/tools.py
goldmann: OK.

concreate/tools.py
dbecvarik: oh right I missed that.

goldmann:
Can someone merge it?

PR #18, opened 2017-09-05
improve support for using directory locations for dependencies

rcernich:
Create distinct ``dependency`` classes to better support directory locations (and distinguish them from git locations).  This includes sprinkling in some OO.


concreate/descriptor.py
```diff
@@ -109,6 +114,58 @@ def _process_labels(self):
...
                                               'value': self.descriptor['description']})
 
+class Dependency(object):
+    def new(dependency):
+        if 'git' in dependency:
+            return GitDependency(dependency['name'], dependency['git']['url'], dependency['git']['ref'])
+        elif 'directory' in dependency:
+            return DirectoryDependency(dependency['name'], dependency['directory'])
+        raise ValueError("Dependency type is not supported: %s" (dependency))
+    new = staticmethod(new)
```
dbecvarik: You can use @staticmethod decorator - it more readable.

concreate/schema/overrides_schema.yaml
```diff
@@ -58,6 +58,13 @@ map:
...
+      - type: map
+        mapping:
+          name: {type: str, required: True}
+          git: 
+              type: map
+              mapping:
+                  url: {type: str, required: True}
+                  ref: {type: str}
+          directory: {type: str, required: False}
+        assert: "val['git'] is not None or val['file'] is not None"
```
dbecvarik: What is val['file'] here? Should it be val['directory'] ?

concreate/descriptor.py
```diff
@@ -109,6 +114,58 @@ def _process_labels(self):
             self.descriptor['labels'].append({'name': 'description',
                                               'value': self.descriptor['description']})
 
+class Dependency(object):
+    def new(dependency):
+        if 'git' in dependency:
```
dbecvarik: If Dependency has both and 'git' and 'directory' defined, git will be used here, is it correct for overrides? I would expect   it vice versa for local development story.

concreate/descriptor.py
```diff
@@ -1,4 +1,6 @@
 import logging
+import subprocess
+import shutil
```
dbecvarik: very minor, but imports should be in alphabetical order.

concreate/descriptor.py
rcernich: Theoretically, it should error if both are present.  As far as merging goes, the merge is done after processing, so the keys should wipe each other out, i.e. cct_module=GitDependency(...) <= cct_module=DirectoryDependency(...), there shouldn't be any overlap, in theory anyway (python isn't my strong suit).

concreate/schema/overrides_schema.yaml
rcernich: Yes, but it doesn't look like assert is working since it's not generating errors. :/

goldmann:
I like this idea/code. Obviously tests need to be fixed and preferably new tests written too.

rcernich:
Related updates to openjdk18-openshift image: https://github.com/jboss-container-images/redhat-openjdk-18-openshift-image/pull/14

rcernich:
Apparently, urlparse is not available during test collection.  I'm guessing this is a problem with the build environment?

rcernich:
This should be good to go.  I migrated the existing tests to use the new Resource classes.  Also, I renamed the Dependency classes to Resource.  These should be used for all "resource" types now.  That said, package repositories are not using Resource, but do share the same download_file() code.

concreate/descriptor.py
```diff
@@ -110,6 +117,164 @@ def _process_labels(self):
...
+        except Exception as ex:
+            # exception is fatal we be logged before Concreate dies
+            raise ConcreateError('Error copying Resource: %s' % (self.name),
+                                 ex)
+
+        self.__verify(target)
+
+        return target
+
+    def __verify(self, target):
```
dbecvarik: please do not use two underscore - i think it has different meaning in python. use only one undescore for 'private' stuff.
for more details search for '__double_leading_underscore' in https://www.python.org/dev/peps/pep-0008/

rcernich: I revisited some of this, but those methods I did want to mark as "private" and I believe two leading underscores is the convention, as it prevents subclasses from being able to override them (name mangling).  I also updated some of the other variables/methods to be inline with the style guide (e.g. _ instead of camel case).

rcernich:
This should be ready to go now.

goldmann:
LGTM, but for final approval I defer here to @dbecvarik.

rcernich:
lemme update the docs

rcernich:
docs have been updated

dbecvarik:
please remove unneeded tests/test_unit_tools.py (contains no tests now) and then we can merge it.

rcernich:
done.

goldmann:
@dbecvarik Are we good to merge it?

PR #20, opened 2017-09-06
Better processing of important labels

### LINKED ISSUE #12: Set summary label to value provided in description key
If the summary label is not defined, its value should be set to value of the
description key. If it's not provided, summary label should not be set.

goldmann:
Labels description and summary are set up correctly based on the description key in the image descriptor.
    
See code for more information.
    
Fixes #12

### DIFF
--- a/.stickler.yml
+++ b/.stickler.yml
@@ -1,2 +1,3 @@
 linters:
-  flake8: {  }
+  flake8:
+    max-line-length: 100
--- a/concreate/descriptor.py
+++ b/concreate/descriptor.py
@@ -44,6 +44,12 @@ class Descriptor(object):
     def items(self):
         return self.descriptor.items()
 
+    def label(self, key):
+        for l in self.descriptor['labels']:
+            if l['name'] == key:
+                return l
+        return None
+
     def process(self):
         """ Prepare descriptor to be used by generating defaults """
         if 'artifacts' in self.descriptor:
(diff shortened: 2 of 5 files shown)


concreate/descriptor.py
```diff
@@ -44,6 +44,12 @@ def __iter__(self):
     def items(self):
         return self.descriptor.items()
 
+    def label(self, key):
+        for l in self.descriptor['labels']:
```
dbecvarik: I would prefer label instead of l here its more readable

concreate/descriptor.py
```diff
@@ -105,10 +111,24 @@ def _process_labels(self):
...
         if "labels" not in self.descriptor:
             self.descriptor['labels'] = []
-        if "description" in self.descriptor:
-            self.descriptor['labels'].append({'name': 'description',
-                                              'value': self.descriptor['description']})
 
+        # The description and maintainer keys available in image descriptor
+        # root are added as labels to the image
+        for key in ['maintainer', 'description']:
+            # If we define the maintainer or description label in the image descriptor
```
dbecvarik: we should decide where maintainer is specified in our tool - i would remove maintainer instruction from concreate and use only labels to be consistent.

.stickler.yml
```diff
@@ -1,2 +1,3 @@
 linters:
-  flake8: {  }
+  flake8:
+    max-line-length: 100
```
dbecvarik: cannot this be read from .flake file? or can we somehow use this linter in any editor like vim/emacs? Maybe we should create some CONTRIBUTE.md describing how to setup local envs.

concreate/descriptor.py
goldmann: I didn't necessary want to interfere with the method name here.

.stickler.yml
goldmann: We will need to have probably both. I don't see any option to make it use `.flake8` file (which is introduced in #21 in fact).

goldmann: I'll add more information about installing the tool as part of #21. I just wanted it here to see how my changes will work. If you want, I can remove it from here and move it to #21.

concreate/descriptor.py
goldmann: The `maintainer` label is specific to Docker container image and it's guidelines. I think it is generic enough to leave it as a first class key.

concreate/descriptor.py
dbecvarik: ok - maybe we can come with better naming later.

concreate/descriptor.py
dbecvarik: Its deprecated in Dockerfile, and image spec from OCI doesn't have it either - i don't think we should implement special key for it. We can generate Dockerfile key but we should have one way of specifying it.

.stickler.yml
dbecvarik: ok

concreate/descriptor.py
goldmann: I can remove the maintainer key.

goldmann: Created: https://github.com/jboss-container-images/concreate/issues/23

dbecvarik: ok, thanks, approving PR. so merge it when you feel it should be merged

goldmann: OCI spec defines "author" as top level key, not label: https://github.com/opencontainers/image-spec/blob/master/config.md#properties

goldmann:
I removed the `maintainer` key. Depending when it'll be approved - I'll update docs.

goldmann:
I'm going to merge this and work on documentation.

PR #21, opened 2017-09-07
Better code formatting

goldmann:


goldmann:
Some checks failed because #20 needs to be merged first.

PR #34, opened 2017-09-11
Introduce generate target

goldmann:
It additionally redefines where we define arguments in CLI.

    concreate --descriptor=image.yaml generate

### DIFF
--- a/concreate/cli.py
+++ b/concreate/cli.py
@@ -42,23 +42,23 @@ class Concreate(object):
                             action='version',
                             help='show version and exit', version=version)
 
-        subparsers = parser.add_subparsers()
+        parser.add_argument('--overrides',
+                            help='path to a file containing overrides')
 
-        parser_build = subparsers.add_parser(
-            'build', help='build container image')
+        parser.add_argument('--target',
+                            default="target",
+                            help="path to directory where to generate sources, \
+                                default: 'target' directory in current working directory")
 
-        parser_build.add_argument('--overrides',
-                                  help='path to a file containing overrides')
+        parser.add_argument('--descriptor',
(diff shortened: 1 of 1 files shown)


dbecvarik:
Just an idea - if ve create build, generate as a positional arguments instead of subparsers we will have better UX. 
then it will be:
condev [opetation] --image.yaml --target --and --all --possible --options
it will also enable you to specify arguments to generate if you are running build command.

goldmann:
I'll rework this to see how it feels. Thanks for suggestion.

goldmann:
I've updated my PoC, here's the output of `--help` command:

```
usage: concreate [-h] [-v] [--version] [--overrides OVERRIDES]
                 [--target TARGET] [--descriptor DESCRIPTOR]
                 {generate,build} [{generate,build} ...]

Dockerfile generator tool

positional arguments:
  {generate,build}      command that should be executed, you can specify
                        multiple commands

optional arguments:
  -h, --help            show this help message and exit
  -v, --verbose         verbose output
  --version             show version and exit
  --overrides OVERRIDES
                        path to a file containing overrides
  --target TARGET       path to directory where to generate sources, default:
                        'target' directory in current working directory
  --descriptor DESCRIPTOR
                        path to image descriptor file, default: image.yaml
```

Example run:

```
concreate generate build --target=target --descriptor=rhel.yaml
```

Please note that `build` "requires" `generate`, so above command could be written just as:

```
concreate build --target=target --descriptor=rhel.yaml
```

dbecvarik:
I'm not 100% sure about build calling target before, but we can change it later. So its LGTM now

goldmann:
You meant 'generate', right?

PR #38, opened 2017-09-12
Move 'user', 'workdir', 'cmd' and 'entrypoint' keys to 'run' section

### LINKED ISSUE #24: Move 'user', 'workdir', 'cmd' and 'entrypoint' keys to 'run' section
Currently we have a top level `user` key. It would make more sense if we move it to `cmd` or `entrypoint` (or another new section we create) so execution and user would be in the same place.

goldmann:
Fixes #24

### DIFF
--- a/concreate/descriptor.py
+++ b/concreate/descriptor.py
@@ -63,6 +63,7 @@ class Descriptor(object):
             self._process_ports()
         if 'dependencies' in self.descriptor:
             self._process_dependencies()
+        self._process_run()
         self._process_labels()
         return self
 
@@ -100,6 +101,15 @@ class Descriptor(object):
                                          execute['execute'])
             if 'user' not in execute:
                 execute['user'] = DEFAULT_USER
+
+    def _process_run(self):
+        """ Make sure the user is set for cmd/entrypoint  """
+
+        if 'run' not in self.descriptor:
+            self.descriptor['run'] = {}
+
+        if 'user' not in self.descriptor['run']:
+            self.descriptor['run']['user'] = DEFAULT_USER
 
     def _process_ports(self):
         """ Generate name attribute for ports """
--- a/concreate/schema/image_schema.yaml
+++ b/concreate/schema/image_schema.yaml
@@ -6,15 +6,6 @@ map:
   version: {type: text, required: True}
   schema_version:  {type: int, required: True}
   description: {type: text}
-  user: {type: text}
-  cmd:
-    type: seq
-    sequence:
-      - type: str
-  entrypoint:
-    type: seq
-    sequence:
-      - type: str
   labels:
     type: seq
     sequence:
@@ -38,7 +29,18 @@ map:
           value: {type: int, required: True}
           expose: {type: bool}
           description: {type: str}
-  workdir: {type: str}
+  run:
+    map:
+      workdir: {type: str}
+      user: {type: text}
+      cmd:
+          type: seq
+          sequence:
+          - type: str
+      entrypoint:
+          type: seq
+          sequence:
+          - type: str
   artifacts:
     seq:
       - type: map
--- a/concreate/templates/template.jinja
+++ b/concreate/templates/template.jinja
@@ -103,17 +103,19 @@ VOLUME ["{{ volume }}"]
 {% endfor %}
 {% endif %}
 
-{%- if workdir %}
+
+USER {{ run['user'] }}
+{%- if run %}
+{%- if 'workdir' in run %}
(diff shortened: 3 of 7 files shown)


goldmann:
This commit outlines one thing that we haven't mentioned yet - should we still make the run/user instruction optional? This may lead to some confusion where the used user is last user that executed any command.

concreate/templates/template.jinja
```diff
@@ -103,17 +103,22 @@ VOLUME ["{{ volume }}"]
...
+{%- if 'workdir' in run %}
 # Specify the working directory
-WORKDIR {{ workdir }}
+WORKDIR {{ run['workdir'] }}
 {% endif %}
 
-USER {{ user }}
+{% if 'user' in run %}
+USER {{ run['user'] }}
+{% endif %}
```
jmtd: > should we still make the run/user instruction optional? This may lead to some confusion where the used user is last user that executed any command.

I think the instruction should remain optional, but should default to root if not specified. There is already an explicit `USER root` earlier on (before packages) in the template. So the default is already root, if no scripts are run. It would be confusing for a `USER` instruction from the scripts section to bleed out to the `CMD` or `ENTRYPOINT` bits at the end. So I'd suggest  adding an `{% else %} USER root` here.

tests/test_dockerfile.py
```diff
@@ -48,10 +46,10 @@ def tearDown(self):
 
     def test_set_cmd_user(self):
         """
-        Ensure that setting a user in the YAML generates a corresponding
+        Ensure that setting 'user' in 'run' section in the YAML generates a corresponding
         USER instruction in the Dockerfile, immediately before the CMD.
         """
-        self.descriptor['user'] = 1347
+        self.descriptor['run']['user'] = 1347
```
jmtd: Oh interesting we default to a non-root UID now. I have no problem with that. This means that `{% if 'user' in run %}` in the jinja will always be true, but it doesn't hurt to have it for robustness.

jmtd:
I don't mind if `/run/user` is optional in the YAML, as long as we continue to set it in the code to a default value, and we should document that too.  The one combination I don't like would be if we made it optional, and removed the default value setting, so the last-run `USER` command (from e.g. a script) bled through.

jmtd:
From local testing we don't generate that last USER command if the value is not set explicitly, so it ends up being root (from the packages section). Adding `{% else %} USER 1347` would be one possible fix

tests/test_dockerfile.py
jmtd: Turns out I was wrong about this.

goldmann:
I think it makes sense to make it `root` when not defined.

goldmann:
@jmtd Reworked.

goldmann:
My proposed fix for the `root` user does not make sense, hold on.

goldmann:
Updated, ready for review. #39 was rebased on top of this.

concreate/descriptor.py
```diff
@@ -96,6 +97,15 @@ def _process_execute(self):
             if 'user' not in execute:
                 execute['user'] = DEFAULT_USER
 
+    def _process_run(self):
+        """ Make sure the user is set for cmd/entrypoint  """
+
+        if 'run' not in self.descriptor:
+            self.descriptor['run'] = {}
+
+        if not self.descriptor['run'].get('user', None):
```
dbecvarik: cannot it be:
'if' user not in self.descriptor['run']:
Also default=None in .get() for dictionary - so you don't need to specify it.

concreate/schema/image_schema.yaml
```diff
@@ -38,7 +29,18 @@ map:
...
           description: {type: str}
-  workdir: {type: str}
+  run:
+    map:
+      workdir: {type: str}
+      user: {type: text}
+      cmd:
+          type: seq
+          sequence:
+          - type: str
```
dbecvarik: this should be text probably - we can have just numeric commands - especially if cmd is used as an argument for entrypoint?

concreate/descriptor.py
goldmann: Sure, it can. Code looked earlier a bit differently and I didn't catch this. Thanks!

concreate/schema/image_schema.yaml
goldmann: This was the case before, I didn't change this. If you think it requires change, please open a ticket.

goldmann:
Updated.

goldmann:
OK, let me update docs, one sec.

PR #39, opened 2017-09-12
Revisit modules section in image descriptor

### LINKED ISSUE #29: Revisit modules section in image descriptor
I think this better describes what this field is used for.

goldmann:
The 'modules' section was reworked. A new 'repositories' section was
added (which replaced the top level 'dependencies') and another
'install' was added to be able to list all modules that need to be
"installed/processed" to generate the image.
    
Fixes #29.

Based on #38.

### DIFF
--- a/concreate/schema/image_schema.yaml
+++ b/concreate/schema/image_schema.yaml
@@ -62,25 +62,27 @@ map:
     seq:
       - {type: str}
   modules:
-    seq:
-      - map:
-          name: {type: str, required: True}
-          version: {type: str}
-  dependencies:
-    seq:
-      - type: map
-        mapping:
-          name: {type: str}
-          git: 
-              type: map
-              mapping:
-                  url: {type: str, required: True}
-                  ref: {type: str}
-          file: {type: str, required: False}
-          url: {type: str, required: False}
-          md5: {type: str}
-          sha1: {type: str}
-          sha256: {type: str}
-          description: {type: str}
-        assert: "val['git'] is not None or val['file'] is not None or val['url] is not None"
+    map:
+      repositories:
+        seq:
+          - type: map
+            mapping:
+              name: {type: str}
+              git:
+                  type: map
+                  mapping:
+                      url: {type: str, required: True}
+                      ref: {type: str}
+              file: {type: str, required: False}
+              url: {type: str, required: False}
+              md5: {type: str}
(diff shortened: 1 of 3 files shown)


goldmann:
I need to rebase properly this PR.

goldmann:
It's done. Once the code is approved, documentation will be added.

concreate/schema/image_schema.yaml
```diff
@@ -62,25 +62,27 @@ map:
     seq:
       - {type: str}
   modules:
-    seq:
-      - map:
-          name: {type: str, required: True}
-          version: {type: str}
-  dependencies:
-    seq:
-      - type: map
```
dbecvarik: use map: only to be consistent with rest of the file

concreate/schema/image_schema.yaml
```diff
@@ -62,25 +62,27 @@ map:
...
-          url: {type: str, required: False}
-          md5: {type: str}
-          sha1: {type: str}
-          sha256: {type: str}
-          description: {type: str}
-        assert: "val['git'] is not None or val['file'] is not None or val['url] is not None"
+    map:
+      repositories:
+        seq:
+          - type: map
```
dbecvarik: again use, map:

dbecvarik:
please make map: vs type:map, mapping: in yaml schema consistent.

goldmann:
This PR needs a bigger rewrite.

goldmann:
Regarding map - I took what Rob proposed in his PR directly and just moved it to new place.

rcernich:
With respect to map, my understanding was that this syntax was required to define the mappings.  The critical piece is that ``git`` is itself a map, i.e. the root map can be one of path, url or git.  I'm no expert at this, so if there's a better way, I'm happy to learn.

dbecvarik:
@rcernich, @goldmann its not a big issue but this https://github.com/jboss-container-images/concreate/blob/develop/concreate/schema/image_schema.yaml#L46-L47
and https://github.com/jboss-container-images/concreate/blob/develop/concreate/schema/image_schema.yaml#L33
are equivalent and we should use only one way in our schema to be consistent. I probably missed this in Rob's PR. So we can fix it now in any subsequent PR.

goldmann:
Closing this PR, I'll open a new one.

PR #42, opened 2017-09-14
Refactor schema definitions

### LINKED ISSUE #32: Refactor schema definitions to reuse common elements
There is not much difference between the three existing schema (image, overrides, module).  It would be nice to define common elements in one place and include them in the specific schema.  This would ease the maintenance burden and would allow the code related to the fields to be common across them (well, this is the way it is, but it prevents missing a schema update should the implementation change).

goldmann:
Schema defeinitions was reworked so that we can easily override
or extend a single definition, depending on the specific need
for particular type of definition (image, overrides, module).

All share the same schema.yaml definition, but modifications are done
in SCHEMA_TYPE.yaml files. Both - extending and overriding is possible.

Fixes #32.


concreate/descriptor.py
```diff
@@ -309,52 +309,3 @@ def _copy_impl(self, target):
         return target
 
 
-def merge_dictionaries(dict1, dict2):
```
dbecvarik: I hope there are no changes here except for moving it to different place - it hard to review otherwise

concreate/schema/module.yaml
```diff
@@ -0,0 +1,7 @@
+map:
```
dbecvarik: Mention, inclusion of common schema in comment.

concreate/schema/image.yaml
```diff
@@ -0,0 +1,2 @@
+map:
```
dbecvarik: Mention, inclusion of common schema in comment.

concreate/schema/overrides.yaml
```diff
@@ -0,0 +1,4 @@
+map:
```
dbecvarik: Mention, inclusion of common schema in comment.

concreate/schema/schema.yaml
```diff
@@ -1,24 +1,20 @@
 # This is a Kwalify Schema for image descriptor.
```
dbecvarik: I would rename this to common.yaml instead of schema.yaml

concreate/tools.py
```diff
@@ -94,27 +95,87 @@ def load_descriptor(descriptor_path, schema_type):
...
+
+    schema_name = '%s.yaml' % schema_type
     schema_path = os.path.join(os.path.dirname(__file__),
                                'schema',
-                               schema_name)
-    if not os.path.exists(schema_path):
-        raise ConcreateError('Cannot locate schema for %s.' % schema_type)
-
+                               'schema.yaml')
+    specific_schema_path = os.path.join(os.path.dirname(__file__),
```
dbecvarik: move this bellow opening schema - so we have schema and specific_schema processing in continuous blocks

concreate/tools.py
```diff
@@ -94,27 +95,87 @@ def load_descriptor(descriptor_path, schema_type):
...
+                                        'schema',
+                                        schema_name)
     schema = {}
```
dbecvarik: maybe rename to common schema?

concreate/descriptor.py
goldmann: There are, I commented on it here: https://github.com/jboss-container-images/concreate/issues/32#issuecomment-329460848

concreate/tools.py
goldmann: We are modifying this later, so it becomes final schema.

concreate/tools.py
```diff
@@ -94,27 +95,87 @@ def load_descriptor(descriptor_path, schema_type):
...
+            dict1[k2] = v2
+        else:
+            if isinstance(v2, list):
+                dict1[k2] = merge_lists(dict1[k2], v2)
+            elif isinstance(v2, dict):
+                dict1[k2] = merge_dictionaries(dict1[k2], v2)
+            else:
+                # if the type is int or strings we override the value from
+                # dict2
+                dict1[k2] = v2
```
goldmann: @dbecvarik Here is the change.

dbecvarik: Isn't this regression for https://github.com/goldmann/concreate/blob/a1a2bea60a7b82917361a2b896267eeceec7c918/concreate/descriptor.py#L78
Especially for name key? So we will override image.yaml `name` with a module.yaml `name`?

goldmann: Good catch!

goldmann: Fixed!

goldmann:
Fixed tests.

goldmann:
Thank you @dbecvarik!

Issue #37, opened 2017-09-12
Document modules


goldmann:
Although we have initial documentation, we still lack documentation for modules.

goldmann:
Closed via https://github.com/jboss-container-images/concreate/pull/45.

Issue #26, opened 2017-09-08
Revisit [repository] section in configuration file


goldmann:
It is not a good idea to set `urls` for repositories globally, because there may be different requirements for different images.

goldmann:
@dbecvarik @rcernich Any comments?

jmtd:
I agree with you on this one, it could causes surprises if you are building images from a variety of different authors and urls set in your global configuration and perhaps forget about it.

rcernich:
Typically, you'd have an option to specify the config.  I think it's fine to look for ``~/.concreate``, but being able to specify ``--config=release.concreate`` would help a lot (or maybe ``--globals``?).  That said, they are referenced in the Dockerfile, so maybe it makes sense to declare them in the yaml.  I don't know what that would look like, as I'm guessing this information is not transportable, other than the raw file name, e.g. ``jboss-rhel-os.repo``, leaving it to the user to provide the actual file.  Is it possible to define multiple repos in a single file?  What happens if the file is empty?  (I'm thinking we could add empty file(s) in the repo and mark them as untracked?)

Also, maybe instead of copying the repos into the image, we could mount a ``repos`` directory into the image and use that instead.  Just an idea.

goldmann:
Thanks Rob for your ideas. These are always valuable!

Let me try to reply one by one.

>  Is it possible to define multiple repos in a single file?
Generally speaking - yes. There is no limit in number of repositories added in a single repo file. The only requirement is that the id (between square bracket) is unique.

This is not true for the current implementation in Concreate. We require that every repo is stored in separate file, the repo id is the same as the file name. This way we have control over what we add, what we enable and what we remove after building the image. This section tried to explain requirements: http://concreate.readthedocs.io/en/develop/configuration.html#repository

> What happens if the file is empty?

YUM/DNF wise - nothing, but current implementation of Concreate will try to enable repo with name defined as the repo file name. This will obviously fails since repo that id will not be found. This failure is fatal.

> Also, maybe instead of copying the repos into the image, we could mount a repos directory into the image and use that instead.

Not supported in container build system we use.

----

My concern is that, although defining them in `.concreate` may be convenient, it can be misleading in cases like:

- One image requires just a subset of "globally" defined repositories
- One image cannot use a specific "global" repository, because it has conflicts with other repository (unlikely, but can happen)
- We build images based on different operating systems (RHEL vs. CentOS vs. Fedora).

Considering above, IMHO we need to define requirements like this:

[...]

PR #45, opened 2017-09-15
Refactoring

### LINKED ISSUE #40: Artifacts section comments
I have a few comments about the `artifacts` section.

1. `file` should generally speaking not be available in the main image descriptor. The reason is that `file` is environment specific, whereas the image descriptor cannot be. It's fine to have `file` in the overrides schema where it makes sense to be defined.
2. I would rename `file` to `path`.
3. How we should define artifacts that are private and we want to only define the name of the artifact? I've received some critic about using the `url` scheme for this use case because a file name is not a URL. Thus we introduced `artifact`.
4. HInts for artifacts were removed. Why?

@rcernich @jmtd @dbecvarik

goldmann:
See individual commit messages for more information.

Fixes #40 #29 #37 #46 #26


goldmann:
Documentation still needs to be updated (I'll do it after actual code review), but I would get feedback on it ASAP so we can finally release Concreate 1.0.

goldmann:
I've actually added initial documentation already.

concreate/descriptor.py
```diff
@@ -116,13 +120,18 @@ def _process_ports(self):
...
             port['name'] = port['value']
 
-    def _process_dependencies(self):
-        """ Generate name attribute for dependencies """
-        dependencies = {}
-        for dependency in self.descriptor['dependencies']:
-            resource = Resource.new(dependency, self.directory)
-            dependencies[resource.name] = resource
-        self.descriptor['dependencies'] = dependencies
+    def _process_module_repositories(self):
```
rcernich: I really think it's better to just define a local relative repository, e.g.:
```yaml
modules:
  repositories:
  - name: java
    path: modules
```
This is explicit and doesn't require any extra magic.  It also allows them to be overridden, if that is desirable, e.g.
```yaml
modules:
  repositories:
  - name: java.platform
    path: modules.rhel
```
where ``modules.rhel`` might look like
```yaml
modules:
  install:
  - module: java.common
```
and could be overridden:
```yaml
modules:
  repositories:
  - name: java.platform
    module: java.centos
```
Which brings up a question, are we using ``name`` for the installed modules, because I think I like what I wrote above (using ``module``), as it's self describing.

concreate/schema/common.yaml
```diff
@@ -51,11 +43,21 @@ map:
...
-        assert: "val['git'] is not None or val['file'] is not None or val['url] is not None"
-  dependencies: *RESOURCES
+        assert: "val['git'] is not None or val['path'] is not None or val['url] is not None"
+  modules:
+    map:
+      repositories: *RESOURCES
+      install:
+        seq:
+          - map:
+              name: {type: str, required: True}
```
rcernich: As noted above, maybe change this to ``module``.

concreate/generator.py
```diff
@@ -117,15 +156,17 @@ def build(self):
...
         # Desired tag of the image
-        tag = "%s:%s" % (self.effective_descriptor['name'], self.effective_descriptor['version'])
+        tag = "%s:%s" % (self.effective_descriptor[
+                         'name'], self.effective_descriptor['version'])
         latest_tag = "%s:latest" % self.effective_descriptor['name']
 
         logger.info("Building %s container image..." % tag)
 
-        ret = subprocess.call(["docker", "build", "-t", tag, "-t", latest_tag, os.path.join(self.target, 'image')])
+        ret = subprocess.call(["docker", "build", "-t", tag,
```
dbecvarik: why not check call if are using ret as rc later?

concreate/descriptor.py
```diff
@@ -25,6 +24,8 @@ def __init__(self, descriptor_path, descriptor_type):
         if descriptor_type == 'image':
             self.check_schema_version()
 
+        self.module_repositories = {}
```
dbecvarik: This doesn't belong to Descriptor class - its not it responsibility to take care repositories implementations - move it to better place (maybe Generator or Repositories class injected to Generator later? )

concreate/generator.py
```diff
@@ -40,13 +44,17 @@ def prepare_modules(self, descriptor=None):
         if not descriptor:
             descriptor = self.descriptor
 
+        modules = descriptor.get('modules', {}).get('install', {})
+
         # If descriptor doesn't requires any module we can start merging descriptors
-        # and fetching artifacts. There ibs nothing left to do except for this
-        if 'modules' not in descriptor:
+        # and fetching artifacts. There is nothing left to do except for this
+        if not modules:
```
dbecvarik: this can be else branch for for cycle after your re factoring which is more readable

concreate/module.py
```diff
@@ -63,13 +63,14 @@ def get_dependencies(descriptor, base_dir):
       descriptor - image descriptor
       base_dir - root directory for dependencies
     """
-    logger.debug("Retrieving dependencies for %s" % (descriptor['name']))
-    if 'dependencies' not in descriptor:
-        logger.debug("No dependencies specified in descriptor")
+    logger.debug("Retrieving module repositories for '%s'" % (descriptor['name']))
+
+    if not descriptor.module_repositories:
```
dbecvarik: this can be else branch again - its more pythonic

concreate/generator.py
goldmann: OK

concreate/generator.py
goldmann: NACK

concreate/generator.py
```diff
@@ -97,14 +107,43 @@ def render_dockerfile(self):
                                   'image',
                                   'Dockerfile')
         with open(dockerfile, 'wb') as f:
-            f.write(template.render(self.effective_descriptor.descriptor).encode('utf-8'))
-        logger.debug("Done")
+            f.write(template.render(
+                self.effective_descriptor.descriptor).encode('utf-8'))
+        logger.debug("Dockerfile rendered")
 
     def prepare_repositories(self):
```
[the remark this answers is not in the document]
goldmann: NACK

concreate/generator.py
```diff
@@ -117,15 +156,17 @@ def build(self):
...
 
         logger.info("Building %s container image..." % tag)
 
-        ret = subprocess.call(["docker", "build", "-t", tag, "-t", latest_tag, os.path.join(self.target, 'image')])
+        ret = subprocess.call(["docker", "build", "-t", tag,
+                               "-t", latest_tag, os.path.join(self.target, 'image')])
 
         if ret == 0:
-            logger.info("Image built and available under following tags: %s and %s" % (tag, latest_tag))
+            logger.info("Image built and available under following tags: %s and %s" % (
```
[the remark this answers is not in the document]
goldmann: Agree.

concreate/descriptor.py
goldmann: Will rewrite this.

concreate/descriptor.py
```diff
@@ -116,13 +120,18 @@ def _process_ports(self):
...
 
-    def _process_dependencies(self):
-        """ Generate name attribute for dependencies """
-        dependencies = {}
-        for dependency in self.descriptor['dependencies']:
-            resource = Resource.new(dependency, self.directory)
-            dependencies[resource.name] = resource
-        self.descriptor['dependencies'] = dependencies
+    def _process_module_repositories(self):
+        local_modules = os.path.join(self.directory, 'modules')
```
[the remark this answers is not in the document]
goldmann: Will change it.

concreate/module.py
goldmann: This will be left as-is.

PR #59, opened 2017-09-19
Fix handling artifacts defined in modules

goldmann:
* Fixed typo for resource description
* Description is displayed when handling of resource fails
* Fixed handling of 'path' resources with cache enabled

### DIFF
--- a/concreate/descriptor.py
+++ b/concreate/descriptor.py
@@ -2,6 +2,7 @@ import logging
 import os
 
 from concreate import DEFAULT_USER, tools
+from concreate.resource import Resource
 from concreate.errors import ConcreateError
 from concreate.version import schema_version
 
@@ -59,6 +60,7 @@ class Descriptor(object):
             self._process_execute()
         if 'ports' in self.descriptor:
             self._process_ports()
+        self._process_artifacts()
         self._process_modules()
         self._process_run()
         self._process_labels()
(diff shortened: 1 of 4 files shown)


concreate/descriptor.py
```diff
@@ -104,6 +106,21 @@ def _process_ports(self):
...
+
+        for artifact in artifacts:
+            if 'name' not in artifact:
+                # This is suboptimal, but at this point we cannot reuse created
+                # resource object. We do create it only to retrieve the name
+                # of the artifact. We do this because we can have different types
+                # of resources (path, url, etc).
+                artifact['name'] = Resource.new(artifact, self.directory).name
+
+        self.descriptor['artifacts'] = artifacts
```
dbecvarik: you are putting Resources classes to descriptor

goldmann: I'm not :)

goldmann: And yes, if you have a better/cleaner idea - please propose one, I couldn't find one.

dbecvarik: whats wrong with old way just - create Resource Object when is needed and not store it in descriptor?
https://github.com/goldmann/concreate/blob/fe7faf84d9e8daa68f253080ea9b179deb0a7182/concreate/generator.py#L82

goldmann: It's dona at different time for modules and main images - thus - we don't have 'name' populated which fails. This was the main reason for this PR.

dbecvarik: So create name here and reuse it later - name should be responsibility of descriptor and resource should only consume it and do not touch it.

goldmann: Yes, we can move responsibility for creating name from resource to descriptor.

goldmann:
@dbecvarik What's the decision here? IIRC on IRC you agreed to approve it, correct?

concreate/descriptor.py
rcernich: Default name generation is resource specific, which is why I put the code in the specific resource objects.  Notice the git resource uses a different algorithm for calculating its default name.  The effect of this is that there ends up being a single "if" block for determining the resource type and all subsequent specialization is done by the class.

Issue #49, opened 2017-09-16
Add support for building images without modules


rcernich:
For simple images, it would be nice to be able to build an image without having to write all the details out into a separate file.  Currently, there are two problems that prevent this:
1. The ``modules`` folder is always copied.  We should always create a modules folder or check to see if it exists before adding the copy command (former seems easier).
2. There needs to be a way to add local scripts into the modules folder.  This seems trickier as we can't just copy the root folder containing the image.yaml, as this can create a recursive copy into itself.  One possible option is to create a special "content" element specifically for image.yaml

goldmann:
It's a nice idea, but we need to discuss in more details how developing of simple images should look. Let's park this idea for now.

rcernich:
I think this also relates to the auto scanning of a local ``modules`` directory, as I suspect that's why we would want to do that (for _simple_ images).

goldmann:
Yah, we discussed it very briefly today but we stopped since we would like to define what a simple image is and how the development workflow should look and what should be required and what not.

rcernich:
I think I've got an example.  Hopefully, I can have something by eod.

rcernich:
This is now working in the referenced PR.  It simply adds a "source" field to image.yaml.  It also expands image.yaml to include support for the "execute" element.

Also, I really don't think we should be automatically adding a local modules repository.  I'll create another issue.

rcernich:
A simple example https://github.com/rcernich/concreate/blob/image-source/image.yaml#L29-L33 which includes all files/folders that are siblings of image.yaml.

dbecvarik:
There is issue with including all siblings - imagine repo where you have your sources and image.yaml in root - we should not copy everything to image as image.yaml was not designed to be a module - if we have simple 
our sources section could be same syntax as .gitignore

``` yaml
source:
  - my_module_dir/*
  - myscript.sh
execute:
  - script: myscript,sh
  - script: my_modudule_dir/foo.sh
```

rcernich:
That's fine, but in the case of the concreate image, it builds the python package and thus needs all the source at the root of the repository.  As far as best practices, I agree, but there's nothing that prevents using "." as the source folder.

As far as filtering goes, if we don't do this, especially with the target/output folder, we run the risk of recursively copying the target folder into the target folder, which could cause space issues on disk.

dbecvarik:
Yes. we can add source . (I would prefer regex syntax here so .* - but that's detail) 
[...]

PR #60, opened 2017-09-19
support source content for image.yaml; add option for specifying configuration file

rcernich:
This fixes a few issues with merging, which was causing exceptions when artifacts were defined in multiple modules.  It also adds support for defining a ``source`` attribute in image.yaml, which effectively turns image.yaml into a full fledged module, with the source content copied into the image's module folder.

I'm working on putting together a builder image for concreate, which will use this new approach for defining an image that does not include any modules.  See issue #49.

This also adds a --config option so you can specify a different configuration file.  This is required so a configuration file can be specified within the builder image.

### DIFF
--- a/concreate/descriptor.py
+++ b/concreate/descriptor.py
@@ -2,8 +2,8 @@ import logging
 import os
 
 from concreate import DEFAULT_USER, tools
-from concreate.resource import Resource
 from concreate.errors import ConcreateError
+from concreate.resource import Resource
 from concreate.version import schema_version
 
 logger = logging.getLogger('concreate')
@@ -17,19 +17,19 @@ class Descriptor(object):
     """
 
     def __init__(self, descriptor_path, descriptor_type):
-        self.directory = os.path.dirname(descriptor_path)
+        self.directory = os.path.abspath(os.path.dirname(descriptor_path))
         self.descriptor = tools.load_descriptor(descriptor_path,
                                                 descriptor_type)
         if descriptor_type == 'image':
-            self.check_schema_version()
+            self.check_schema_version(descriptor_path)
 
-    def check_schema_version(self):
+    def check_schema_version(self, descriptor_path):
         """ Check supported schema version """
         if self.descriptor['schema_version'] != schema_version:
-            raise ConcreateError("Schema version: '%s' is not supported by current version."
+            raise ConcreateError("%s: Schema version '%s' is not supported by current version."
                                  " This version supports schema version: '%s' only."
                                  " To build this image please install concreate version: '%s'"
-                                 % (self.descriptor['schema_version'],
+                                 % (descriptor_path, self.descriptor['schema_version'],
                                     schema_version,
                                     self.descriptor['schema_version']))
 
@@ -56,11 +56,12 @@ class Descriptor(object):
 
     def process(self):
         """ Prepare descriptor to be used by generating defaults """
+        if 'artifacts' in self.descriptor:
+            self._process_artifacts()
         if 'execute' in self.descriptor:
             self._process_execute()
         if 'ports' in self.descriptor:
             self._process_ports()
-        self._process_artifacts()
         self._process_modules()
         self._process_run()
         self._process_labels()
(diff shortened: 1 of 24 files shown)


goldmann:
FYI: Fixing descriptor merging was reviewed and is going to be merged as part of #59.

rcernich:
This is ready to go.  I'm not sure why you couldn't create the resource objects when the descriptor was processed.  I actually think this is required to support relative paths.

The image.yaml creates an openshift s2i image that can be used to generate the image source.  There's probably some more work to do here, since the output image contains all the concreate stuff, but it serves its purpose.  The template creates two build configs, one that builds the source image and one that builds the actual image.  There are two secrets that are required.  concreate-config must contain a .concreate configuration file.  package-repos must contain .repo files required for package installation.

dbecvarik:
This is changing a lot of unrelated stuff in concreate for our current roadmap. Can we have this PR separated to one containing OCP builder image only? 
An then create separate issue for rest with proper milestone settings and proper discussion?

goldmann:
Agree with @dbecvarik.

rcernich:
I can remove the dict_ids bit from the merge code.  Other than that, all the other changes are required for the OCP builder image.  Those include:
* Adding an option for specifying a .concreate configuration file
* Building an image with no modules
* Resolving repo files using relative paths to the .concreate file

As for the ``descriptor['artifacts']`` changes, I'm not sure why you changed those in the first place.  I continually have problems running the openjdk refactor without them.  Also, I don't see how relative paths work without them too.

concreate/descriptor.py
```diff
@@ -73,13 +74,25 @@ def merge(self, descriptor):
...
 
+    def _process_artifacts(self):
+        """ Processes descriptor artifacts section and generate default
+        value 'name' for each artifact which doesnt have 'name' specified.
+        """
+        artifacts = {}
+        for artifact in self.descriptor['artifacts']:
+            resource = Resource.new(artifact, self.directory)
+            artifacts[resource.name] = resource
+        self.descriptor['artifacts'] = artifacts
```
dbecvarik: descriptor should not contain any object different to dictionary - we should build proper objects for everything or nothing.

concreate/generator.py
```diff
@@ -44,6 +45,35 @@ def prepare_modules(self, descriptor=None):
         """
         if not descriptor:
             descriptor = self.descriptor
+            if 'source' in descriptor:
+                # copy image source
+                ignored_paths = set()
```
dbecvarik: I would prefer to go with "white listing" parts of the modules - blacklisting is hard to maintain and make concreate a very hard to put to existing repositories (a lot of changes can introduce a noisy commits adding path to blacklist)

image/added/concreate/repos/jboss-rhel-rhscl.repo
```diff
@@ -0,0 +1,6 @@
+[jboss-rhel-rhscl]
+name=Red Hat RHEL7 SCL repo
+baseurl=
```
dbecvarik: why have any repo file with empty base url? this doesn't make sense for me.

concreate/schema/image.yaml
```diff
@@ -1,3 +1,10 @@
 # To be merged with common.yaml
 map:
   from: {type: str, required: True}
+  version: {type: text, required: False}
```
dbecvarik: cannot this be in common schema?

concreate/tools.py
```diff
@@ -16,10 +16,14 @@
...
 
-def parse_cfg():
+def init_cfg(config_file):
+    config_file = os.path.expanduser(config_file)
+    logger.info("Using concreate configuration file: %s" % (config_file))
     cp = configparser.ConfigParser()
-    cp.read(os.path.expanduser('~/.concreate'))
-    return cp._sections
+    cp.read(config_file)
+    global cfg
```
dbecvarik: this is not needed - modules can be used as "singletons"

concreate/descriptor.py
rcernich: Agreed, but that's quite a large change and descriptor is an object already.

concreate/generator.py
rcernich: That's fine, but I'm not sure how that would work for the example here.  Also, it's very problematic to not filter ./target/ since this causes a recursive copy into itself, which could blow up the disk.

image/added/concreate/repos/jboss-rhel-rhscl.repo
rcernich: These are there simply as placeholders.  I wasn't sure what to put in these to make them work ootb, so i left them blank.

concreate/schema/image.yaml
rcernich: probably

concreate/tools.py
rcernich: This supports --config=myconfig.ini

concreate/schema/image.yaml
rcernich: I've addressed this through an additional commit.

image/added/concreate/repos/jboss-rhel-rhscl.repo
rcernich: As these are in the process of being removed, I'd like to hold off removing it here until the other work is complete.

concreate/generator.py
rcernich: I've removed filtering of local module folders through an additional commit.

concreate/tools.py
dbecvarik: global keyword is not required. you can do import tools. then tools.cfg is behaving like singleton.

rcernich: global is required as ``cfg`` is defined in this package (i.e. this is in the tools package itself).

dbecvarik:
Some of the ideas were implement in a different way -closing as it will probably not be updated

PR #61, opened 2017-09-20
Add support for specifying custom tags for the build target

### LINKED ISSUE #33: Allow to specify a custom tag for the 'build' target
Currently to tags are added:

1. `$IMAGE_NAME/$IMAGE_VERSION`
2. `$IMAGE_NAME/latest`

We should be able to override it by specifying the tag (with `--tag` parameter) added to the `build` target.

goldmann:
It's now possible to override default tags using the --build-tag
parameter. Multiple tags can be specified.

Fixes #33.

### DIFF
--- a/concreate/generator.py
+++ b/concreate/generator.py
@@ -20,6 +20,7 @@ logger = logging.getLogger('concreate')
 class Generator(object):
 
     def __init__(self, descriptor_path, target, overrides):
+
         self.descriptor = Descriptor(descriptor_path, 'image').process()
         self.target = target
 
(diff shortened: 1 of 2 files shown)


concreate/generator.py
```diff
@@ -19,11 +19,13 @@
 
 class Generator(object):
 
-    def __init__(self, descriptor_path, target, overrides):
-        self.descriptor = Descriptor(descriptor_path, 'image').process()
-        self.target = target
+    def __init__(self, args):
```
dbecvarik: I don't like this change - it hides what Generator constructor needs to be passed in, who know what part of args it takes?

concreate/generator.py
```diff
@@ -19,11 +19,13 @@
...
 
-        if overrides:
+        self.descriptor = Descriptor(args.descriptor, 'image').process()
+        self.target = args.target
+        self.build_tags = args.build_tags
```
dbecvarik: maybe this should not be part of constructor but build can take tags as an argument?

concreate/generator.py
```diff
@@ -155,23 +157,33 @@ def build(self):
...
+        tags = ["%s:%s" % (self.descriptor['name'], self.descriptor['version']), "%s:latest" % self.descriptor['name']]
+
+        # Custom tags for the container image
+        if self.build_tags:
+            logger.debug("Custom tags for container image specified: %s" % ", ".join(self.build_tags))
+            tags = self.build_tags
+
+        for tag in tags:
+            cmd.append("-t")
+            cmd.append(tag)
```
dbecvarik: cmd.append('-t %s' % tag) 
or just lambda one liner?

concreate/generator.py
```diff
@@ -155,23 +157,33 @@ def build(self):
...
 
         logger.info("Building %s container image..." % tag)
 
+        cmd.append(os.path.join(self.target, 'image'))
+
         try:
-            subprocess.check_call(["docker", "build", "-t", tag,
-                                   "-t", latest_tag, os.path.join(self.target, 'image')])
+            subprocess.check_call(cmd)
```
dbecvarik: Log constructed command in debug level.

goldmann: Good idea.

concreate/generator.py
goldmann: The reasoning was to not endlessly extend the constructor, but I can revert it.

dbecvarik: see comments bellow about build(), maybe you don't need tags in constructor at all.

PR #68, opened 2017-09-25
Created custom builder engine for Docker and OSBS

dbecvarik:


concreate/builder.py
```diff
@@ -0,0 +1,37 @@
...
+      build_engine - a build engine to use to build image
+      image_dir - path where image sources are generated
+    """
+
+    def __new__(cls, build_engine, image_dir):
+        print build_engine
+        if cls is Builder:
+            if 'docker' == build_engine:
+                # import is delayed until here to prevent circular import error
+                from concreate.builders.docker import DockerBuilder
```
goldmann: You could possibly do:

```python
from concreate.builders.docker import DockerBuilder as BuilderImpl
```

and then just run one `super` call.

concreate/cli.py
```diff
@@ -62,6 +64,10 @@ def run(self):
                             default="image.yaml",
                             help="path to image descriptor file, default: image.yaml")
 
+        parser.add_argument('--build-engine',
+                            default='docker',
+                            help='An engine used to build an image [docker,osbs]')
+
```
goldmann: This should be in the "build" section of arguments.

concreate/cli.py
```diff
@@ -45,6 +47,10 @@ def run(self):
         build_group = parser.add_argument_group(
             'build', "Arguments valid for the 'build' target")
 
+        build_group.add_argument('--build-engine',
+                                 default='docker',
+                                 help='An engine used to build an image [docker,osbs]')
```
goldmann: You should specify these using this: https://docs.python.org/3/library/argparse.html#choices

concreate/schema/common.yaml
```diff
@@ -66,3 +66,9 @@ map:
       install:
         seq:
           - {type: str}
+  osbs:
```
goldmann: Shouldn't this be part of the image.yaml only? I don't think we will be building modules as images.

concreate/builders/osbs.py
```diff
@@ -0,0 +1,203 @@
...
+
+    def prepare(self, descriptor):
+        """Prepares dist-git repository for OSBS build."""
+
+        repository_key = descriptor.get('osbs', {}).get('repository', {})
+        repository = repository_key.get('name')
+        branch = repository_key.get('branch')
+
+        if not (repository and branch):
+            raise ConcreateError("Osbs builder needs repostiory and branch provided!")
```
goldmann: "OSBS"

concreate/builders/osbs.py
```diff
@@ -0,0 +1,203 @@
...
+
+        self.dist_git_dir = os.path.join(os.path.expanduser('~/.concreate.d'),
```
goldmann: I'm a bit unsure about this directory name, but I think it could be left.

concreate/schema/common.yaml
dbecvarik: what about overrides?

concreate/builders/osbs.py
dbecvarik: yes. Im not sure too - but i didn't want to change .concreate to directory in this PR

concreate/schema/common.yaml
goldmann: In overrides too. For such case, where we override it - the "base" image.yaml should not contain the `osbs` key even.

concreate/builders/docker.py
```diff
@@ -0,0 +1,46 @@
...
+        This can be changed by specifying the tags in CLI using --build-tags option.
+
+        Args:
+          build_tags - a list of image tags
+        """
+        tags = build_args.build_tags
+        cmd = ["docker", "build"]
+
+        # Custom tags for the container image
+        logger.debug("Custom tags for container image specified: %s" %
```
goldmann: This should be reworded. After your refactoring these can be default or custom tags.

concreate/schema/common.yaml
dbecvarik: I think it even make sense for module - we can have centos/fedora/rhel flavor of module which describes only where image is builded instead of overrides.

concreate/builder.py
```diff
@@ -0,0 +1,36 @@
...
+            return super(Builder, cls).__new__(BuilderImpl)
+
+    def __init__(self, build_engine, target):
+        self.target = target
+
+    def prepare(self, descriptor):
+        # we dont require prepare to be implemented by builder
+        pass
+
+    def build(self, build_tags=[]):
```
goldmann: We should log what type of builder is used in info log level.

concreate/builders/osbs.py
```diff
@@ -0,0 +1,206 @@
...
+        with Chdir(self.dist_git_dir):
+            self.dist_git.add()
+
+            if self.dist_git.stage_modified():
+                self.dist_git.commit()
+                self.dist_git.push()
+            else:
+                logger.info("No changes made to the code, committing skipped")
+
+            logger.info("Executing container build on OSBS...")
```
goldmann: We should NOT execute a build in case we don't push changes to dist-git repo. In other words, when we answer "N" when asked about pushing to repo - we cannot kick a build and builder should immediately exit.

concreate/cli.py
```diff
@@ -45,10 +47,21 @@ def run(self):
         build_group = parser.add_argument_group(
             'build', "Arguments valid for the 'build' target")
 
+        build_group.add_argument('--build-engine',
+                                 default='docker',
+                                 choices=['docker', 'osbs'],
+                                 help='An engine used to build an image [docker,osbs]')
```
goldmann: No need to repeat choices, these are displayed by default.

goldmann: And probably reworded: "Engine used to build the image".

concreate/cli.py
```diff
@@ -45,10 +47,21 @@ def run(self):
...
                                  dest='build_tags',
                                  action='append',
-                                 help='tag to assign to the built image, can be used multiple times')
+                                 help='Tag to assign to the built image, can be used multiple times')
+
+        build_group.add_argument('--build-osbs-release',
+                                 dest='build_osbs_release',
+                                 action='store_true',
+                                 default=False,
+                                 help='Do a proper OSBS build not scratch only.')
```
goldmann: I would probably reword this and drop "scratch". Hard to define what "proper" is :) "Execute release build". We can and maybe should even highlight that this is important switch.

concreate/builders/docker.py
```diff
@@ -0,0 +1,46 @@
...
+        This can be changed by specifying the tags in CLI using --build-tags option.
+
+        Args:
+          build_tags - a list of image tags
+        """
+        tags = build_args.build_tags
+        cmd = ["docker", "build"]
+
+        # Custom tags for the container image
+        logger.debug("Building image with a custom tags: '%s'" %
```
goldmann: You don't know whether these tags are custom or not. Please change it.

docs/build.rst
```diff
@@ -0,0 +1,44 @@
+Building image
+================
+
+The main purpose of concreate is abstraction of building containers image from common source
+using custom builder engines.
+
+Concreate now support following builder engines:
+
+* ``Docker`` -- build the image using docker build command and it default option
+* ``OSBS`` -- build the image using OSBS service
```
goldmann: Add link to OSBS upstream project.

concreate/cli.py
```diff
@@ -45,10 +47,21 @@ def run(self):
...
+        build_group.add_argument('--build-engine',
+                                 default='docker',
+                                 choices=['docker', 'osbs'],
+                                 help='An engine used to build an image [docker,osbs]')
+
         build_group.add_argument('--build-tag',
                                  dest='build_tags',
                                  action='append',
-                                 help='tag to assign to the built image, can be used multiple times')
+                                 help='Tag to assign to the built image, can be used multiple times')
```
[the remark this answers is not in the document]
goldmann: You are mixing upper and lower case in CLI help sentences. Default `--help` uses lower case, this is where I standardized my help descriptions.

docs/build.rst
```diff
@@ -0,0 +1,44 @@
...
+
+* ``--build-engine`` -- a builder engine to use ``osbs`` or ``docker``
+* ``--build-tag`` -- an image tag used for build image (can be specified multiple times)
+* ``--build-osbs-release`` -- perform a osbs release build
+
+
+``Docker build``
+^^^^^^^^^^^^^^^^
+
+This is the default way how to build an container image. The image is build using ``docker``.
```
goldmann: `docker build`.

PR #72, opened 2017-09-27
Added basic tech preview support

dbecvarik:


concreate/cli.py
```diff
@@ -61,7 +61,12 @@ def run(self):
                                  dest='build_osbs_release',
                                  action='store_true',
                                  default=False,
-                                 help='execute OSBS release build.')
+                                 help='execute OSBS release build')
+
+        parser.add_argument('--tech-preview',
```
goldmann: Shouldn't this be `--build-tech-preview`? It's a build flag.

concreate/cli.py
```diff
@@ -61,7 +61,12 @@ def run(self):
...
+                            action='store_true',
+                            default=False,
```
goldmann: False is default by default :)

concreate/cli.py
dbecvarik: i'm not sure its basically generate option - but this should be hidden in future - so Im not sure here

concreate/cli.py
dbecvarik: good point

concreate/cli.py
goldmann: You even added it to the build documentation...

docs/build.rst
```diff
@@ -23,7 +23,7 @@ You can execute image build by running:
 * ``--build-engine`` -- a builder engine to use ``osbs`` or ``docker``
 * ``--build-tag`` -- an image tag used for build image (can be specified multiple times)
 * ``--build-osbs-release`` -- perform a OSBS release build
-
+* ``--build-tech-preview`` -- updates JBOSS_IMAGE_NAME variable to contain '-tech-preview' suffix in family part of the image name
```
goldmann: Do not mention here that variable. Important is only that the name, before slash will be suffixed.

PR #93, opened 2017-10-11
Created a local development documentation, fixed overrides for module repos

dbecvarik:


docs/local_development.rst
```diff
@@ -0,0 +1,45 @@
...
+
+.. code:: bash
+
+  $ git clone https://github.com/jboss-openshift/cct_module.git /home/user/repo/cct_module
+
+2. Create override.yaml next to the image.yaml. override.yaml should look like:
+
+.. code:: yaml
+
+  modules:
```
goldmann: Misses `schema_version` key.

docs/local_development.rst
```diff
@@ -0,0 +1,45 @@
...
+      - name: s2i-common
+
+
+And we are in a need of updating s2i-common module. To achieve this one of the possible way is to:
+
+1. Clone cct_module to your workstation to ``~/repo/cct_module``
+
+.. code:: bash
+
+  $ git clone https://github.com/jboss-openshift/cct_module.git /home/user/repo/cct_module
```
goldmann: This should be intended byt another 2 spaces IIRC.

docs/local_development.rst
```diff
@@ -0,0 +1,45 @@
...
+
+2. Create override.yaml next to the image.yaml. override.yaml should look like:
+
+.. code:: yaml
+
+  modules:
+    repositories:
+      - path: "/home/user/repo/cct_module"
```
goldmann: This example does not work, name for git is the whole url whereas for path is the whole path. This means that it will not be overridden. Am I correct?

concreate/descriptor.py
```diff
@@ -127,14 +127,23 @@ def _process_modules(self):
...
 
+        for repository in repositories:
+            name = ""
+            if 'git' in repository:
+                name = repository['git']['url']
+            elif 'path' in repository:
+                name = repository['path']
+            elif 'url' in repository:
+                name = repository['url']
+            repository['name'] = os.path.splitext(os.path.basename(name))[0]
```
goldmann: If `name` is already set, you will overwrite it by executing this.

goldmann:
I gave it a shot and generally speaking - its not working.

When set up similar environment, as described in the attached documentation, we try to override - the `git` resource is not overridden. Instead, the `path` module is added to the module and effectively ignored.

```
{'repositories': [{'path': '/home/goldmann/git/public/cct_module', 'git': {'url': 'https://github.com/jboss-openshift/cct_module.git', 'ref': 'master'}, 'name': 'main'}]}
```

P.S. I defined the name for the both modules, as the code doesn't work either with this regard, soo inline comment.

docs/local_development.rst
dbecvarik: Where exactly?

docs/local_development.rst
dbecvarik: no - we are doing splitext for name

PR #98, opened 2017-10-12
Add missing volume information in schema

### LINKED ISSUE #91: include 'volumes' in the schema definition
Volumes is included in the templates, but not the schema.

```
ERROR    ('Cannot validate schema: image.yaml', SchemaError(msg='Schema validation failed:
 - Key 'volumes' was not defined. Path: ''.'))
```

We should be able to add the following to image.yaml
```
volumes: 
    - "/var/run/artemis" 
```

goldmann:
Fixes #91

### DIFF
--- a/concreate/schema/common.yaml
+++ b/concreate/schema/common.yaml
@@ -72,3 +72,8 @@ map:
         map:
           name: {type: str}
           branch: {type: str}
+  volumes:
+    seq:
+      - map:
+          name: {type: str}
+          path: {type: str, required: True}
--- a/docs/descriptor.rst
+++ b/docs/descriptor.rst
@@ -355,8 +355,11 @@ In case you want to define volumes for your image, just use the ``volumes`` sect
 .. code:: yaml
 
     volumes:
-        - "/opt/eap/standalone"
+        - name: "volume.eap"
+          path: "/opt/eap/standalone"
 
+.. note::
 
+    The ``name`` key is optional. If not specfied the value of ``path`` key will be used.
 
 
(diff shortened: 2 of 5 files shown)


concreate/schema/common.yaml
```diff
@@ -72,3 +72,6 @@ map:
         map:
           name: {type: str}
           branch: {type: str}
+  volumes:
+    seq:
+      - {type: str}
```
dbecvarik: if we do it this way, volumes cannot be overridden, only appended, do we really want this?

goldmann: So, this was something I was thinking about and I went with this approach for a simple reason: this is the way it is described in docs and tests, but it was just missing in the schema. I could update it to sequence of map though.

Another possibility is to actually merge two lists: `list1 + list2` in our merging algorithm.

dbecvarik: how to merge? we are doing list1 + list2, issue is if you want to remove/change volume. Because with plain list we will very soon get bug that we cannt override them similar to repositories.

goldmann: Oh, yes, sorry, I was too quick.

So, what to do with this? I'm actually cool with either option.

docs/descriptor.rst
```diff
@@ -355,8 +355,11 @@ In case you want to define volumes for your image, just use the ``volumes`` sect
 .. code:: yaml
 
     volumes:
-        - "/opt/eap/standalone"
+        - name: "volume.eap"
+          path: "/opt/eap/standalone"
 
+.. note::
 
+    The ``name`` key is optional. If not specfied the vlaue of ``path`` fwill be used.
```
dbecvarik: typo - vlaue -> value

PR #105, opened 2017-10-13
Fix modules overrides

### LINKED ISSUE #92: package repositories override feature, doesn't remove the original repositories
When using the --overrides flag, it builds with the repositories from the original image.yaml, and the override_image.yaml.

image.yaml (upstream)
```
packages:
      repositories:
          - centos-base
      install:
          - openssl
```
override.yaml (downstream)
```
packages:
      repositories:
          - jboss-os
      install:
          - openssl  
```
Results in 
`Step 7 : RUN yum install -y --disablerepo=\* --enablerepo=jboss-rhel-os --enablerepo=centos-base openssl     && yum clean all &&     rpm -q  openssl`
RHEL does not like the base urls for the centos repos.  It should only contain the repositories from the override.yaml

dbecvarik:
Partially fixes #92 
WE should introduce a module versioning and enable proper module overrides.

### DIFF
--- a/concreate/generator.py
+++ b/concreate/generator.py
@@ -66,9 +66,7 @@ class Generator(object):
         logger.info("Handling modules...")
 
         for module in modules:
-            version = None
-            if 'version' in module:
-                version = module['version']
+            version = module.get('version', None)
 
             req_module = copy_module_to_target(module['name'],
                                                version,
--- a/docs/overrides.rst
+++ b/docs/overrides.rst
@@ -1,0 +1,130 @@
+Overrides
+=========
+
+During an image life cycle there can be a need to do a slightly tweaked builds - using different base images, injecting newer libraries etc. We want to support such scenarios without a need of duplicating whole image sources. To achieve this Concreate supports overrides mechanism for its image descriptor. You can override almost anything in image descriptor. The overrides are based on overrides descriptor - a yaml file containing changes to image descriptor.
+
+To use an override descriptor you need to pass ``--overides <PATH>`` argument to a concreate.
+
+**Example**: To use overrides.yaml  file located in current working directory run:
+
+.. code:: bash
+
+	  $ concreate build --overrides=overrides.yaml
+
+How overrides works
+-------------------
+
+Concreate is using `YAML <http://yaml.org/>`_ format for its descriptors. Overrides in concreate works on `YAML node <http://www.yaml.org/spec/1.2/spec.html#id2764044>`_ level.
+
+
+Scalar nodes
+^^^^^^^^^^^^
+Scalar nodes are easy to overrides, if Concreate finds any scalar node in a overrides descriptor it update its value in image descriptor with the overridden one.
+
+**Example**: Overriding scalar node:
+
+*image descriptor*
+
+.. code:: yaml
+
+	  schema_version: 1
+	  name: "dummy/example"
+	  version: "0.1"
+	  from: "busybox:latest"
+
+*overrides descriptor*
+
+.. code:: yaml
+
+	  schema_version: 1
+	  from: "fedora:latest"
+
+*overridden image descriptor*
+
+.. code:: yaml
+
+	  schema_version: 1
+	  name: "dummy/example"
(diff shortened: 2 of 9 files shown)


concreate/generator.py
```diff
@@ -66,9 +66,7 @@ def prepare_modules(self, descriptor=None):
         logger.info("Handling modules...")
 
         for module in modules:
-            version = None
-            if 'version' in module:
-                version = module['version']
+            version = module.get('version', None)
```
goldmann: Nice.

dbecvarik: nice catch, fixing

dbecvarik:
Fixes #92

docs/overrides.rst
```diff
@@ -0,0 +1,130 @@
...
+
+How overrides works
+-------------------
+
+Concreate is using `YAML <http://yaml.org/>`_ format for its descriptors. Overrides in concreate works on `YAML node <http://www.yaml.org/spec/1.2/spec.html#id2764044>`_ level.
+
+
+Scalar nodes
+^^^^^^^^^^^^
+Scalar nodes are easy to overrides, if Concreate finds any scalar node in a overrides descriptor it update its value in image descriptor with the overridden one.
```
goldmann: typo: "easy to override"

goldmann: typo: "an overrides descriptor"

goldmann: typo: "it updates""

docs/overrides.rst
```diff
@@ -0,0 +1,130 @@
...
+
+**Example**: Overriding scalar node:
+
+*image descriptor*
```
goldmann: Maybe instead of this use: `image.yaml` and `override.yaml`? This will play nicely together with the example at the top.

docs/overrides.rst
```diff
@@ -0,0 +1,130 @@
...
+.. code:: yaml
+
+	  schema_version: 1
+	  name: "dummy/example"
+	  version: "0.1"
+	  from: "fedora:latest"
+
+Sequence nodes
+^^^^^^^^^^^^^^
+Sequence nodes are little bit tricky, if they're representing plain arrays, we cannot easily override any value so Concreate is just merging arrays from image and override descriptors together. This can have unexpected results, please see `corresponding issue. <https://github.com/jboss-container-images/concreate/issues/106>`_
```
goldmann: I would probably add a top level section: known issues, or even create a separate page for known issues and link to it from here.

docs/overrides.rst
```diff
@@ -0,0 +1,130 @@
...
+
+.. code:: yaml
+
+	  schema_version: 1
+	  name: "dummy/example"
+	  version: "0.1"
+	  from: "busybox:latest"
+	  run:
+	    cmd:
+  	    - "bar"
```
goldmann: Is it really added as the first argument? That's a bit weird.

docs/overrides.rst
```diff
@@ -0,0 +1,130 @@
...
+	    - "echo"
+	    - "foo"
+
+Mapping nodes
+^^^^^^^^^^^^^
+Mappings are merged via *name* key. If Concreate is overriding an mapping or array of mappings it tries to find a *name* key in mapping and use and identification of mapping. If two *name* keys matches, all keys of the mapping are updated.
```
goldmann: `name` should be backslashed twice IMHO

docs/overrides.rst
dbecvarik: its not - overrides are taken first - so everything is added to overrides in the code base.

docs/overrides.rst
dbecvarik: If I do this, how will I call the third one? :)

goldmann: `result.yaml`?

docs/overrides.rst
dbecvarik: I've highlighted it little - but i prefer to have it tight with the docs, so its easy to spot

PR #109, opened 2017-10-16
Updated URLResource to try to use url directly if cacher miss occurs.

### LINKED ISSUE #77: Should try original url/path when downloading resource from cache fails
If an artifact is missing from the cache, the build will fail without trying to download the resource from the specified url/path.  This may be a misunderstanding regarding how the cache is meant to be used.  The understanding being the cache is used for performance, as opposed to restricting downloads to known, pre-verified artifacts.

dbecvarik:
fixes #77

### DIFF
--- a/concreate/resource.py
+++ b/concreate/resource.py
@@ -137,9 +137,10 @@ class Resource(object):
                         .replace('#hash#', self.checksums[algorithm]))
         return url
 
-    def _download_file(self, url, destination):
+    def _download_file(self, url, destination, use_cache=True):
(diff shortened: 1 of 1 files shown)


concreate/resource.py
```diff
@@ -137,9 +137,10 @@ def __substitute_cache_url(self, url):
                         .replace('#hash#', self.checksums[algorithm]))
         return url
 
-    def _download_file(self, url, destination):
+    def _download_file(self, url, destination, use_cache=True):
```
goldmann: Smart move, +1

PR #114, opened 2017-10-18
Extend dockerfile generation tests

dbecvarik:


goldmann:
Is it possible to add some description what the specific test does? Right now it's a bit unclear. I mean here these specific lines in pytest.

PR #119, opened 2017-11-02
Migrated concreate descriptor to be object based

dbecvarik:
Now every important part of the descriptor is treated as object. With
this we can easily extend our schema and use object api instead of
plain dictionaries inside code.

This also enables to have multiple schemas per object type, so we can
easily introduce incompatible changes to our schema without breaking
old images. Each object is resposinble to handling its schema and
converting it to compatible object state.


concreate/generator.py
```diff
@@ -20,7 +20,7 @@ class Generator(object):
 
     def __init__(self, descriptor_path, target, overrides):
 
-        self.descriptor = Descriptor(descriptor_path, 'image').process()
+        self.descriptor = Image(tools.load_descriptor(descriptor_path))
```
goldmann: I would say `self.descriptor` is a misleading name for an field now.

concreate/generator.py
```diff
@@ -96,7 +96,7 @@ def prepare_artifacts(self):
 
     def override(self, overrides_path):
         logger.info("Using overrides file from '%s'." % overrides_path)
-        descriptor = Descriptor(overrides_path, 'overrides').process()
+        descriptor = Overrides(tools.load_descriptor(overrides_path))
```
goldmann: Same here, I would not call it descriptor anymore.

concreate/module.py
```diff
@@ -95,7 +96,7 @@ class Module(Descriptor):
     descriptor_path: A path to module descriptor file.
     """
     def __init__(self, descriptor_path):
-        self.descriptor = Descriptor(descriptor_path, 'module').process()
+        self.descriptor = Module(tools.load_descriptor(descriptor_path))
```
goldmann: And here too.

concreate/generator.py
dbecvarik: Do you have any suggestion? Its still a descriptor - only difference is that it is divided into containers not plain dictionaries.

dbecvarik:
@goldmann can you please review again?

PR #125, opened 2017-11-09
Implemented --test-steps-url parameter to fetch customs steps

dbecvarik:


concreate/cli.py
```diff
@@ -53,6 +53,11 @@ def parse(self):
                                 action='store_true',
                                 help='Run @wip tests only')
 
+        test_group.add_argument('--test-steps-url',
+                                default='https://github.com/jboss-container-images/'
+                                'concreate-test-steps.git',
```
goldmann: This line breaking is more than weird.

dbecvarik: its how you handle long string in python - multiple strings are concatenated together

PR #126, opened 2017-11-09
moved --tag to be common option not build group related one.

dbecvarik:


concreate/cli.py
```diff
@@ -64,12 +64,8 @@ def parse(self):
...
                                  dest='build_tags',
                                  action='append',
-                                 help='tag to assign to the built image, can be used multiple times')
-
-        build_group.add_argument('--tag',
-                                 dest='tags',
-                                 action='append',
-                                 help='tag used to build/test the image, can be used multiple times')
+                                 help='tag to assign to the built image, '
+                                 'can be used multiple times')
```
goldmann: This line breaking is too eager.

dbecvarik: its not, 101 characters is more than 100 allowed for us

goldmann: This is I don't like it. At all.

goldmann:
There is a conflict.

concreate/cli.py
dbecvarik: Can you suggest better breaking?

goldmann: Sure, not breaking it :)

dbecvarik: I can do something like:
``` python
        build_group.add_argument('--build-tag',
                                 dest='build_tags',
                                 action='append',
                                 help=('tag to assign to the built image, '
                                       'can be used multiple times'))
```

which is maybe indented better?

PR #128, opened 2017-11-16
Implemented smarter arrays handling and overriding

dbecvarik:
The documentation is not changed, because feature is stil considered
internal and in a preparation state

### DIFF
--- a/concreate/descriptor.py
+++ b/concreate/descriptor.py
@@ -195,12 +195,59 @@ class Packages(Descriptor):
         map:
           repositories:
             seq:
-              - {type: str}
+              - {type: any}
           install:
             seq:
-            - {type: str}""")]
+              - {type: str}"""),
+                        yaml.safe_load("""
+        map:
+          repositories:
+            seq:
+              - {type: any}
+          packages:
+            seq:
+              - {type: any}""")]
 
         super(Packages, self).__init__(descriptor)
+
+        if 'install' in self.descriptor:
+            self.descriptor['packages'] = []
+            for pkg in self.descriptor['install']:
+                self.descriptor['packages'].append(ListItem(pkg))
+        elif 'packages' in self.descriptor:
+            pkgs = self.descriptor['packages'][:]
+            self.descriptor['packages'] = []
+            for pkg in pkgs:
+                self.descriptor['packages'].append(ListItem(pkg))
+
+        if 'repositories' in self.descriptor:
+            repos = self.descriptor['repositories'][:]
+            self.descriptor['repositories'] = []
+            for repo in repos:
+                self.descriptor['repositories'].append(ListItem(repo))
+
+
+class ListItem(Descriptor):
+    def __init__(self, descriptor):
+        self.schemas = [yaml.safe_load("""
+        map:
+          name: {type: str, required: True}
+          state:
+            type: str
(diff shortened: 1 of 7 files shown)


concreate/descriptor.py
```diff
@@ -195,13 +195,60 @@ def __init__(self, descriptor):
...
+            pkgs = self.descriptor['packages'][:]
+            self.descriptor['packages'] = []
+            for pkg in pkgs:
+                self.descriptor['packages'].append(ListItem(pkg))
+
+        if 'repositories' in self.descriptor:
+            repos = self.descriptor['repositories'][:]
+            self.descriptor['repositories'] = []
+            for repo in repos:
+                self.descriptor['repositories'].append(ListItem(repo))
```
goldmann: Above code looks like some repetition, but I have no idea how to write it better/clearer.

concreate/descriptor.py
```diff
@@ -195,13 +195,60 @@ def __init__(self, descriptor):
...
+            seq:
+              - {type: any}""")]
 
         super(Packages, self).__init__(descriptor)
 
+        if 'install' in self.descriptor:
+            self.descriptor['packages'] = []
+            for pkg in self.descriptor['install']:
+                self.descriptor['packages'].append(ListItem(pkg))
+        elif 'packages' in self.descriptor:
```
goldmann: This is the "old" way, right? Maybe print a warning?

concreate/templates/template.jinja
```diff
@@ -46,16 +46,16 @@ EXPOSE {%- for port in helper.ports(ports) %} {{ port }}{% endfor %}
...
 # Add custom repo files
 COPY repos/*.repo /etc/yum.repos.d/
 {% endif %}
 
 # Install required RPMs and ensure that the packages were installed
-RUN yum install -y {% if additional_repos -%} --disablerepo=\* {%- for repo in additional_repos %} --enablerepo={{ repo }}{% endfor %}{% endif %} {%- for package in packages.install %} {{ package }}{% endfor %} \
+RUN yum install -y {% if additional_repos -%} --disablerepo=\* {%- for repo in additional_repos %} --enablerepo={{ repo }}{% endfor %}{% endif %} {%- for package in helper.filter_state(packages.packages, 'present') %} {{ package }}{% endfor %} \
     && yum clean all && \
-    rpm -q {% for package in packages.install %} {{ package }}{% endfor %}
+    rpm -q {% for package in helper.filter_state(packages.packages, 'present') %} {{ package }}{% endfor %}
```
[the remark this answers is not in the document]
dbecvarik: its universal helper for ListItem objects, so that's the reason why its taking arguments.

concreate/descriptor.py
dbecvarik: the idea is, that both approaches are valid. So you can still write simple way, if you dont need to override.

concreate/templates/template.jinja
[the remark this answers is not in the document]
goldmann: Are we really interested in objects that are not 'present'?

concreate/descriptor.py
goldmann: Then we need to have it documented. Very well documented.

dbecvarik:
this is outdated now and needs bigger refactorign, closing.

PR #143, opened 2017-11-27
test documentation overhaul

jmtd:
Some changes to the test documentation page based in part on some older documentation from before we used concreate.

### DIFF
--- a/docs/test.rst
+++ b/docs/test.rst
@@ -1,9 +1,11 @@
-Testing image
-=============
+Testing images
+==============
 
-Concreate is able to run `behave <https://pythonhosted.org/behave/>`_ based test for the image. We suggest to read behave documentation before reading this chapter.
+Concreate is able to run `behave <https://pythonhosted.org/behave/>`_ based
+test for images. We suggest you read the Behave documentation before reading
+this chapter.
 
-Image can be tested by running:
+An image can be tested by running:
 
 .. code:: bash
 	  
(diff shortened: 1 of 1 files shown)


docs/test.rst
```diff
@@ -1,57 +1,85 @@
...
 	  
 	  $ concreate test
 
 **Test options**
 
-* ``--test-wip`` -- run only tests tagged with ``@wip`` tag.
+* ``--test-wip`` -- only run tests tagged with the ``@wip`` tag.
 * ``--test-steps-url`` -- a git repository url containing `steps <https://pythonhosted.org/behave/tutorial.html#python-step-implementations>`_ for tests.
-* ``--tag`` -- an image tag used to determine image which will be tests. It can be specified multiple times but only the first occurence is honored.
+* ``--tag @sometag`` --  only run tests tagged with e.g. ``@sometag``. Only the first occurrence of this argument is honoured.
```
dbecvarik: IIRC this is not intended, it works the way, that it still run tests according the image name from image.yaml and --tag is used to determine the tag of image which should be tested.

docs/test.rst
```diff
@@ -1,57 +1,85 @@
...
-   This is very special behavior used mainly in development. It servers purpose you want to limit test to be run to a subset you are working on. To achieve this you should mark your test scenario with ``@wip`` tag and run ``concreate test --test-wip``
+   This is very special behavior used mainly in development. Its purpose is to
+   to limit the tests to be run to a subset you are working on. To achieve this
+   you should mark your in-development test scenarios with the ``@wip`` tag and
+   run ``concreate test --test-wip``.
 
 3. `Custom tags`
 
-   You can test an particular image using via ``--tag`` option. To run test for image 'foo' you can run ``concreate test --tag foo``
+   You can restrict tests to those for a particular image using the ``--tag``
```
dbecvarik: see comment above

docs/test.rst
jmtd: Hi David, yes reading the source it seems if you are running "test" then --tag gets ignored. So I'm not sure what is desired behaviour here.  What should --tag do in "test" case?

dbecvarik: --tag should choose proper image to tests. So if you create image named 'foo' and you want run tests for that image, you use --tag foo and it will run the test against the image named 'foo'. Test tags are constructed from image yaml.

Issue #144, opened 2017-11-28
cannot override repositories in modules element


rcernich:
There seems to be a general issue with the merge capabilities since moving to Descriptor objects.

dbecvarik:
Fixed via #147  + added tests

PR #145, opened 2017-11-28
fix merge functionality for Descriptor objects

rcernich:

Addresses #144.

### DIFF
--- a/concreate/tools.py
+++ b/concreate/tools.py
@@ -7,6 +7,7 @@ import os
 import shutil
 import yaml
 
+from concreate.descriptor import Descriptor
 from concreate.errors import ConcreateError
 
 
@@ -63,6 +64,9 @@ def merge_dictionaries(dict1, dict2, kwalify_schema=False):
 
     Return merged dictionaries
     """
+    if dict2 is None:
+        return dict1
+
     for k2, v2 in dict2.items():
         if k2 not in dict1:
             dict1[k2] = v2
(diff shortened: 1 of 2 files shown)


dbecvarik:
I've already created different PR, with different aproach but similar outcome. 
https://github.com/jboss-container-images/concreate/pull/128

My approach is little different, it enables you fully override (not only concatenate list).

its based on assumption, that if we have list objects like:
``` yaml
list:
   - foo
   - bar
```
I enable alternative syntax like:
``` yaml
list:
   - name: foo # you dont need state:present here as this si default
   - name: bar
     state: absent
   - name: baz
     state: present
```
I support both syntax for all of our list object, so you can still use plain list form, but if you need an override (in overrides.yaml) you can use more complex syntax

goldmann:
@rcernich @dbecvarik Can you guys talk to each other and decide on the approach?

concreate/tools.py
```diff
@@ -100,7 +107,8 @@ def merge_lists(list1, list2):
             else:
                 for v1 in list1_dicts:
                     if v2['name'] == v1['name']:
-                        merge_dictionaries(v1, v2)
+                        if isinstance(v1, dict) and isinstance(v2, dict):
```
dbecvarik: Name here is to be sure if we have list of dictionaries, that we are merging proper ones - thiw will just pair them in order, Im not sure if this works correctly.

dbecvarik:
I've got inspired and implemented necessary parts to enable proper merging of descriptor based object in #147

dbecvarik:
ITs probably not relevant anymore

PR #147, opened 2017-11-28
Fixed merging of dictionaries

dbecvarik:
Dictionaries can be also Descriptor classes now.


goldmann:
Merged, make sure to add it to develop too!

PR #148, opened 2017-11-29
Implemented backward compatibilty for local modules overrides

dbecvarik:


concreate/generator.py
```diff
@@ -19,9 +19,16 @@
...
-        self.descriptor = Image(tools.load_descriptor(descriptor_path),
-                                os.path.dirname(descriptor_path))
+        descriptor = tools.load_descriptor(descriptor_path)
+
+        # if there is a local modules directory and no modules are defined
+        # we will inject it for a backward compatibility
+        local_mod_path = os.path.join(os.path.abspath(os.path.dirname(descriptor_path)), 'modules')
+        if os.path.exists(local_mod_path):
+            modules = descriptor.get('modules', [])
+            if not modules.get('repositories'):
```
goldmann: This will fail. Array does not have "get" method.

concreate/generator.py
```diff
@@ -19,9 +19,16 @@
...
-
-        self.descriptor = Image(tools.load_descriptor(descriptor_path),
-                                os.path.dirname(descriptor_path))
+        descriptor = tools.load_descriptor(descriptor_path)
+
+        # if there is a local modules directory and no modules are defined
+        # we will inject it for a backward compatibility
+        local_mod_path = os.path.join(os.path.abspath(os.path.dirname(descriptor_path)), 'modules')
+        if os.path.exists(local_mod_path):
+            modules = descriptor.get('modules', [])
```
goldmann: It should be dict.

concreate/generator.py
goldmann: Additionally, if you switch to dict, this means that this will throw exception.

concreate/generator.py
```diff
@@ -19,9 +19,16 @@
...
-        self.descriptor = Image(tools.load_descriptor(descriptor_path),
-                                os.path.dirname(descriptor_path))
+        descriptor = tools.load_descriptor(descriptor_path)
+
+        # if there is a local modules directory and no modules are defined
+        # we will inject it for a backward compatibility
+        local_mod_path = os.path.join(os.path.abspath(os.path.dirname(descriptor_path)), 'modules')
+        if os.path.exists(local_mod_path) and 'modules' in descriptor:
+            modules = descriptor.get('modules')
+            if not modules.get('repositories'):
```
goldmann: Will it not throw KeyError? Should be: `if not modules.get('repositories', null):`

goldmann: Or `{}`

dbecvarik: nope, None is default IIRC

goldmann: Nope: https://docs.python.org/2/library/stdtypes.html#typesmapping

get(key[, default])
Return the value for key if key is in the dictionary, else default. If default is not given, it defaults to None, so that this method never raises a KeyError.

goldmann: You are right in fact. I read the docs too fast, sorry!

PR #162, opened 2017-12-14
Modules required by other modules are now correctly discovered

dbecvarik:


concreate/module.py
```diff
@@ -74,6 +74,7 @@ def get_dependencies(descriptor, base_dir):
 
     for repo in module_repositories:
         repo.copy(base_dir)
+        discover_modules(os.path.join(base_dir, repo['name']))
```
goldmann: Someone forgot about this? :D

dbecvarik: maybe..... :)

PR #167, opened 2018-01-08
Refatored descriptor.py to multiple files for easier readability

dbecvarik:


goldmann:
Very nice and indeed readable, good job!

PR #175, opened 2018-01-15
Enabled --overrides option to contain inlined YAML

dbecvarik:


concreate/tools.py
```diff
@@ -32,18 +32,23 @@ def cleanup(target):
...
     Returns descriptor as a dictionary
     """
-    logger.debug("Loading descriptor from path '%s'." % descriptor_path)
+    if not os.path.exists(descriptor):
+        logger.info("Descriptor path '%s' doesn't exists, trying to parse it directly."
+                    % descriptor)
+        try:
+            return yaml.safe_load(descriptor)
+        except Exception:
+            raise ConcreateError('Cannot load descriptor.')
```
goldmann: Probably try to make it more understandable of what happened and why it could fail. We need to be clear about this.

concreate/tools.py
```diff
@@ -32,18 +32,23 @@ def cleanup(target):
...
 
-    if not os.path.exists(descriptor_path):
-        raise ConcreateError('Cannot find provided descriptor file')
+    logger.debug("Loading descriptor from path '%s'." % descriptor)
```
goldmann: This may or may not be true. Needs rewording.

dbecvarik: Is it True all the time, in the case YAML is specified, code will not reach this part, or am I missing some part?

Issue #161, opened 2017-12-14
Validate image.yaml before running


dbecvarik:
When we define any object multiple time in yaml file, it now generates same number of descriptor objects, we should print warning about that, because some values can be covered.

dbecvarik:
related to  https://github.com/concrt/concreate/issues/155

PR #179, opened 2018-01-17
Fixed Image schema validaiton

dbecvarik:
It seems that iamge_schema got overriden by override schema, so name
key was not mandatory.
I've fixed it by exposing only deep copy of image_schema.


dbecvarik:
Fixed #161

PR #187, opened 2018-02-01
Image repositories are not disabled by yum anymore

dbecvarik:
Concreate was chaged in a way, that existing repositories in image
are not disabled during build phase. This enables easy community
images development.

### DIFF
--- a/concreate/templates/template.jinja
+++ b/concreate/templates/template.jinja
@@ -49,7 +49,12 @@ USER root
 {% if packages and packages.install %}
 {% if packages.repositories %}
 # Add custom repo files
-COPY repos/*.repo /etc/yum.repos.d/
+COPY \
+{% for repo in packages.repositories %}
+    repos/{{ repo.name }}.repo
+{% endfor %}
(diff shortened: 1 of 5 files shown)


dbecvarik:
Do not merge, I found a bug I need to fix during community image streams builds.

dbecvarik:
this should be ready to merge now

concreate/templates/template.jinja
```diff
@@ -47,19 +47,19 @@ EXPOSE {%- for port in helper.ports(ports) %} {{ port }}{% endfor %}
 USER root
 
 {% if packages and packages.install %}
-{% if additional_repos %}
+{% if packages.repositories %}
 # Add custom repo files
 COPY repos/*.repo /etc/yum.repos.d/
```
goldmann: We probably should add selectively files here, just like we do it at removal time, so nothing we do not expect will not be added to container and then NOT removed.

dbecvarik: good catch, thanks

Issue #204, opened 2018-03-22
Use --pull=true option for builds


rcernich:
This causes lots of problems building locally, when an old base image is used.

dbecvarik:
fiixed by #208

PR #208, opened 2018-04-17
Implemented --build-pull option

dbecvarik:
if you run Cekit with --build-pull it asks builder to fetch latest
base image


dbecvarik:
fixes #204

PR #233, opened 2018-05-21
Fixed where os.path.isdir failed for git modules repositories

dbecvarik:


dbecvarik:
This issue was not catched by unit tests as we mock os.path.isdir https://github.com/cekit/cekit/blob/develop/tests/test_unit_resource.py#L25

Issue #106, opened 2017-10-16
Overriding/merging of sequence nodes can lead to weird behavior.


dbecvarik:
In current conreate, if we have sequence nodes of plain values like:

``` yaml
run:
  cmd: 
    - "foo"
    - "bar"
```
and any module which define this `cmd` key too, all values are appended and can result in non expected state.

We probably should consider moving everything to mappings syntax, so we can easily override/append values.

Keys, affected by this are:
* packages/repositories
* packages/install
* run/cmd
* run/entrypoint

rcernich:
Ordering should be a high priority to ensure module dependencies are applied before the module depending on them.  For example, if module A requires module B, module B should be applied to the image before module A.  A more complex example:

```yaml
name: module.a
modules:
  install
  - name: module.c
```

```yaml
name: module.b
modules:
  install
  - name: module.c
```

```yaml
name: image
modules:
  install:
  - name: module.a
  - name: module.b
```

When the image is processed the modules should be applied in the following order: module.c, module.a, module.b.

I assume this is how things work today, I just want to make sure it does work this way.

dbecvarik:
yes. this is how its implemented now. This issue is about naked "lists" like 
``` yaml
run:
  cmd:
    - 'foo'
    - 'bar'
```
Which cannot be easily extended/overridden.

rcernich:
I think things like this would have specialized merge code.  For the example above, I don't think it makes sense to merge the command list at all.  Instead, one would most like override the entire cmd field.

PR #235, opened 2018-05-29
update overrides bug link

elmiko:
the link to issue #106 in the documentation pointed to an old repository that no longer exists. this change refreshes the link to its current location.


goldmann:
Thank you!

PR #238, opened 2018-06-04
Fix a typo

nalind:
When invoking buildah with _pull=True, use the --pull-always option.


goldmann:
@dbecvarik It looks like the test is broken too :)

https://github.com/cekit/cekit/blob/develop/tests/test_builder.py#L167

nalind:
Whoops, didn't catch the in-tree test.

goldmann:
@nalind Thank you!

PR #239, opened 2018-06-04
All execute section should now be ordered properly.

dbecvarik:
Merging was changed to be bottom up with merging execute blocks before
their parent. This should carry whole subrtrees up in proper places.

### DIFF
--- a/cekit/descriptor/__init__.py
+++ b/cekit/descriptor/__init__.py
@@ -10,7 +10,7 @@ from cekit.descriptor.run import Run
 from cekit.descriptor.packages import Packages
 from cekit.descriptor.packages import Repository
 from cekit.descriptor.modules import Modules
-from cekit.descriptor.execute import Execute
+from cekit.descriptor.execute import Execute, ExecuteContainer
 from cekit.descriptor.image import Image
 from cekit.descriptor.module import Module
 from cekit.descriptor.overrides import Overrides
--- a/cekit/descriptor/base.py
+++ b/cekit/descriptor/base.py
@@ -102,6 +102,9 @@ class Descriptor(collections.MutableMapping):
     def items(self):
         return self._descriptor.items()
 
+    def __repr__(self):
+        return "%s" % self._descriptor
+
     def get(self, k, default=None):
         return self._descriptor.get(k, default)
 
@@ -135,10 +138,10 @@ def _merge_descriptors(desc1, desc2):
         if k2 not in desc1:
             desc1[k2] = v2
         else:
-            if isinstance(v2, list):
+            if isinstance(v2, Descriptor):
+                desc1[k2].merge(v2)
+            elif isinstance(v2, list):
                 desc1[k2] = _merge_lists(desc1[k2], v2)
-            elif isinstance(v2, Descriptor):
-                desc1[k2] = _merge_descriptors(desc1[k2], v2)
     return desc1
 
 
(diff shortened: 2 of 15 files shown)


rcernich:
I think it would be better to preprocess the list of modules first, i.e. flatten the included modules into a sequential list, then perform the merges.  This fix address the issue of script execution, but what about labels, environment variables, rpm installs, etc.?  Collecting and ordering all the modules will address all of these fringe cases.

dbecvarik:
We don't have other stuff which requires strict ordering.  If we add something like this we can just refactor ExecutionContainer to multipurpose container and hold it. 
We cant easily flatten to sequential list as dependency model is a tree. So we will end up in creating tree and traversing through it similar to this recursion.

rcernich:
> We don't have other stuff which requires strict ordering. If we add something like this we can just refactor ExecutionContainer to multipurpose container and hold it.

This is not true.  A module might explicitly override an environment variable's value.  The concept of modules is that whatever metadata they apply to the image happens after any metadata changes applied by its dependencies.  This issue specifically applies to script ordering, but the same concept applies to all aspects of a module.

> We cant easily flatten to sequential list as dependency model is a tree. So we will end up in creating tree and traversing through it similar to this recursion.

This can be accomplished by creating a list of the modules to be included.  The basic algo:

```python
def flattenDependencies(imageOrModule, modules):
  if imageOrModule in modules:
    return:
  for module in imageOrModule.modules.install:
    flattenDependencies(module, modules)
  modules.append(imageOrModule)

def processImage(image):
  modules = []
  flattenDependencies(image, modules)
  image.modules = modules
  # merge
```

dbecvarik:
> This is not true. A module might explicitly override an environment variable's value. The concept of modules is that whatever metadata they apply to the image happens after any metadata changes applied by its dependencies. This issue specifically applies to script ordering, but the same concept applies to all aspects of a module.

Yes and that's the reason why this recursion used there and we are merging from bottom up. This assures that overriding of variables etc resolves correctly even if same variable is placed in the multiple branches of the tree.  
Maybe I didn't get your suggestion properly, can we talk about it probably on team meeting on Wednesday?

dbecvarik:
> This can be accomplished by creating a list of the modules to be included. The basic algo:
> 
> def flattenDependencies(imageOrModule, modules):
>   if imageOrModule in modules:
>     return:
>   for module in imageOrModule.modules.install:
>     flattenDependencies(module, modules)
>   modules.append(imageOrModule)
> 
> def processImage(image):
>   modules = []
>   flattenDependencies(image, modules)
>   image.modules = modules

I was trying to think about this code and this is exactly how it works. It seems very similar for what we do now.  Currently we work the way, that we do depth-first tree traversal and we are merging from the bottom (ordering is probably completely same as creating the linked list, right?).
This means that we are merging child module to the parent. This is important as if we have any variable or label or any dictionary defined in the parent and child, the value from parent is kept. Also this means, that merging of arrays which requires strict ordering (execute only for now) is tricky as we need to merge it in the proper place (if you merge child into parent, you need to place executes of *child* before the parent ones as they need to be executed first).
This is not an issue for all other stuff as we are changing order of variables and label in the code (sorting them alphabetically).

Your suggestion would be nice if we will be merging parent to child (reversed than we do now) and we will not keep duplicate values but override them.  This will probably solve the issue with merging arrays as they can be easily appended. The issue with this approach is that it means to completely re-factor merging code and almost every unittest which is testing merging as the algorithm will be reversed.

This will of course result in a nicer code, but its some work. Maybe its good idea to do it now, before we get wider adoption of the tool and these changes will be harder to do.

PR #252, opened 2018-06-11
Revamped docs

dbecvarik:


goldmann:
Looks very good, there are some generation warnings, it would be good to silence these and there are a few pages that are missing content, like main page of operation guide. Additionally main page has wrong link.

PR #255, opened 2018-06-13
Implemented overrides chaining

dbecvarik:
Cekit now accepts multiple overrides via command line options both
--overrides and --overrides-file can be chained together.

Overrides are resolved in order they appear so values from the
override which appeared first take precedence to later ones.

### DIFF
--- a/docs/overrides.rst
+++ b/docs/overrides.rst
@@ -16,9 +16,18 @@ To use an override descriptor you need to pass ``--overides-file`` argument to a
 
 .. code:: bash
 
-	  $ cekit build --overrides "{'labels': [{'name': 'foo', 'value': 'overriden'}]}"
+	  $ cekit build --overrides "{'labels': [{'name': 'foo', 'value': 'overridden'}]}"
 
+Overrides Chaining
+------------------
 
+You can even specify multiple overrides. They are resolved in order they appear on command line. This means that values from *first override specified overrides all values from later overrides*.
+
+**Example**: If you run following command, label `foo` will be set to `bar`.
+
+.. code:: bash
+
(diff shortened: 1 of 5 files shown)


docs/overrides.rst
```diff
@@ -16,9 +16,18 @@ To use an override descriptor you need to pass ``--overides-file`` argument to a
...
 
-	  $ cekit build --overrides "{'labels': [{'name': 'foo', 'value': 'overriden'}]}"
+	  $ cekit build --overrides "{'labels': [{'name': 'foo', 'value': 'overridden'}]}"
 
+Overrides Chaining
+------------------
 
+You can even specify multiple overrides. They are resolved in order they appear on command line. This means that values from *first override specified overrides all values from later overrides*.
+
+**Example**: If you run following command, label `foo` will be set to `bar`.
```
goldmann: I would expect it to be set to `baz`. Can you confirm that this is just a typo?

iankko: +1. IMHO if we want multiple overrides to be supported, the final form should be taken from the last provided one. If the first is the one, which takes effect, what's the point of trying to redefine the same value again?

But having the last one applied, suppose you have complex (in git repo present already) `override-cp.yaml` file, and you know, you want to change just one line of it. If the last one is the honoured one, you would do:

```
{concreate,cekit} build -v --build-engine=... --overrides=override-cp.yaml --overrides=my-custom-one-line-overrides.yaml
```

It seems kinda natural to me, the last specified option is the applied one. IMHO having the first one being honoured is prone to many errors / mistakes => new rebuilds needed.

Just my 0.02.

PR #259, opened 2018-07-12
--redhat flag now set cacher url for ce cacher

dbecvarik:


cekit/cli.py
```diff
@@ -178,6 +178,10 @@ def run(self):
 
             if self.args.redhat:
                 tools.cfg['common']['redhat'] = True
+
+            if tools.cfg['common']['redhat'] and 'cache_url' not in tools.cfg['common']:
+                tools.cfg['common']['cache_url'] = \
+                    'http://ce-cacher.usersys.redhat.com/get?#algorithm#=#hash#'
```
goldmann: Internal URL in upstream code?

dbecvarik: seems like its common practice https://pagure.io/odcs/blob/master/f/client/contrib/odcs#_24

goldmann: I'm not saying no, just this would be precedence in our code since we haven't done anything like this so far. This, of course, makes things much easier for internal users.

dbecvarik: ok lets ask @jmtd and @rcernich for opinions?

goldmann: Agree.

dbecvarik: This is only default. You can still override it by defining in ~/.cekit/config as you can do now..

rcernich: And anybody running cekit who doesn't update their default will be trying to hit this url.

dbecvarik: Yep you are probably rigth. Closing this. I'm still thinking more and more of moving this redhat related stuff into cekit-redhat-plugin rpm or something more feasible;

PR #260, opened 2018-07-12
[CLOUD-2698] add usage: label to image schema

jmtd:
usage: label has turned up in our base images and we need to be able
to override it.

https://issues.jboss.org/browse/CLOUD-2698


goldmann:
I'm -1 for it. It should be a simple label, not an additional key in the descriptor,

jmtd:
Oh sorry of course. My mistake

PR #272, opened 2018-07-26
Container help pages

jmtd:
This is a bit WIP still but raising PR for feedback. My own list of things still todo

 - [x] <s>should we always generate help.md? Or make it toggleable. If so, default to on?</s>flag/config option to include generate docs in image
   - [x] + key in ~/.cekit config
 - [x] document the help.md generation in cekit's docs
 - [x] <s>intersection with usage label? if usage label is not present, should we add a label pointing at the location of the help.md in the image?</s> punt
 - [x] <s>markdown or asciidoc? for asciidoc, we could use tables for some things (similar to our old autogenerated docs). But are there plans for other systems to consume these pages, and if so, what would they accept?</s> markdown for now
 - [ ] support user-provided template file...
    - [x] specified in config file
    - [ ] specified on command line?
    - [ ] specified in image yaml
 - [x] (determine and write) relevant tests

Example of output docs: https://github.com/jmtd/redhat-openjdk-18-openshift-image/blob/help-pages/target/image/help.md

### DIFF
--- a/cekit/cli.py
+++ b/cekit/cli.py
@@ -192,16 +192,15 @@ class Cekit(object):
             tools.cfg['common']['redhat'] = True
         if self.args.work_dir:
             tools.cfg['common']['work_dir'] = self.args.work_dir
+
+        addhelp = tools.cfg.get('doc',{}).get('addhelp', False)
         if bool == type(self.args.addhelp):
-            tools.cfg['common']['addhelp'] = self.args.addhelp
-        else: # NoneType
-            if not 'addhelp' in tools.cfg['common']:
-                tools.cfg['common']['addhelp'] = False
+            addhelp = self.args.addhelp
 
         # We need to construct Generator first, because we need overrides
         # merged in
         params = {
-            'addhelp': tools.cfg['common']['addhelp'],
+            'addhelp': addhelp,
             'redhat':  tools.cfg['common']['redhat'],
         }
         self.generator = Generator(self.args.descriptor,
--- a/cekit/cli.py
+++ b/cekit/cli.py
@@ -141,6 +141,20 @@ class Cekit(object):
                             default="image.yaml",
                             help="path to image descriptor file, default: image.yaml")
 
+        addhelp_group = parser.add_mutually_exclusive_group()
+
+        addhelp_group.add_argument('--add-help',
+                                   dest='addhelp',
+                                   action='store_const',
+                                   const=True,
+                                   help="Include generate help files in the image")
+
+        addhelp_group.add_argument('--no-add-help',
+                                   dest='addhelp',
+                                   action='store_const',
+                                   const=False,
+                                   help="Do not include generate help files in the image")
+
         parser.add_argument('commands',
                             nargs='+',
                             choices=['generate', 'build', 'test'],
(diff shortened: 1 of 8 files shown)


cekit/cli.py
```diff
@@ -141,6 +141,11 @@ def parse(self):
                             default="image.yaml",
                             help="path to image descriptor file, default: image.yaml")
 
+        parser.add_argument('--add-help',
```
dbecvarik: please make it also key in ``~/.cekit/config`` so it  can be set for whole tool.

cekit/templates/template.jinja
```diff
@@ -126,6 +126,10 @@ WORKDIR {{ run['workdir'] }}
 ENTRYPOINT {{ helper.cmd(run['entrypoint']) }}
 {% endif %}
 
+{% if addhelp %}
```
dbecvarik: i would like to move this at the end as I would like to keep ENTRYPOINT and CMD  together.

cekit/cli.py
jmtd: np (added to my TODO checklist)

cekit/templates/template.jinja
jmtd: np - will fix with my next push

jmtd:
I've localled implemented config file support but the matrix of options is now sufficiently complex (config or not, command line or not, command line inverse or not) that I need to write the tests to make sure all combinations are correct.

jmtd:
small bits in addition to the other TODO list

- [ ] bug: config file "addhelp = True" not working yet
- [ ] enhancement: dynamically generate tests (smaller, scale)
- [ ] enhancement: catch and fail if --no-add-help and --add-help both specified
  - [ ] and test

cekit/tools.py
```diff
@@ -28,6 +28,8 @@ def get_cfg(config_path):
     cfg['common'] = cfg.get('common', {})
     cfg['common']['work_dir'] = cfg.get('common').get('work_dir', '~/.cekit')
     cfg['common']['redhat'] = cfg.get('common', {}).get('redhat', False)
+    if cp.has_option('common', 'addhelp'):
```
dbecvarik: maybe it would be nice to create `[doc]` section. You can have `addhelp` there and also probably a way to custom template.

jmtd: That sounds like a good idea. I'll implement the custom template stuff and then rework both to use [doc].

cekit/tools.py
```diff
@@ -28,8 +28,8 @@ def get_cfg(config_path):
     cfg['common'] = cfg.get('common', {})
     cfg['common']['work_dir'] = cfg.get('common').get('work_dir', '~/.cekit')
     cfg['common']['redhat'] = cfg.get('common', {}).get('redhat', False)
-    if cp.has_option('common', 'addhelp'):
-        cfg['common']['addhelp'] = cp.getboolean('common', 'addhelp')
+    if 'doc' in cp and cp.has_option('doc', 'addhelp'):
```
jmtd: dang this doesn't work

jmtd: …except it does. Further testing (by me) needed

dbecvarik:
LGTM, merging. Thanks!

PR #274, opened 2018-07-31
Implemented a way to remove any key

dbecvarik:
If you set any attribute to in YAML to None(~) it will be removed

### DIFF
--- a/cekit/cache/artifact.py
+++ b/cekit/cache/artifact.py
@@ -34,14 +34,16 @@ class ArtifactCache():
         return self._get_cache()
 
     def add(self, artifact):
-        if not artifact.checksums:
+        if 'md5' not in artifact and \
+           'sha256' not in artifact and \
+           'sha512' not in artifact:
             raise ValueError('Cannot cache Artifact without checksum')
 
         artifact_id = str(uuid.uuid4())
 
         for alg in SUPPORTED_HASH_ALGORITHMS:
-            if alg in artifact.checksums:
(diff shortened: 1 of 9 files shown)


cekit/cache/artifact.py
```diff
@@ -34,14 +34,16 @@ def list(self):
         return self._get_cache()
 
     def add(self, artifact):
-        if not artifact.checksums:
+        if 'md5' not in artifact and \
+           'sha256' not in artifact and \
+           'sha512' not in artifact:
```
jmtd: what do you think of `if set(SUPPORTED_HASH_ALGORITHMS) & set(artifact):` here?

cekit/descriptor/base.py
```diff
@@ -126,6 +126,26 @@ def process_defaults(self):
...
+        if isinstance(self, Descriptor):
+            _remove_none_keys(self)
+        else:
+            # it means it list
+            for desc in self:
+                _remove_none_keys(desc)
+
+
+def _remove_none_keys(desc):
+    for key in {k: v for k, v in desc.items()}:
```
jmtd: this is a shallow clone so you can delete keys, I think? this might work: `for key in dict(desc):`

cekit/cache/artifact.py
jmtd: Or equivalently `set(SUPPORTED_HASH_ALGORITHMS).intersection(set(artifact))`

PR #276, opened 2018-07-31
Split Cekit.run into run, configure

jmtd:
Split the run() method into a configure() method and a run()
method that calls it.

This means we can write tests for the configure() code without
invoking run() and generating Dockerfiles etc.

(Review tip: try ignoring whitespace changes with "git show HEAD -w")

### DIFF
--- a/cekit/cli.py
+++ b/cekit/cli.py
@@ -160,7 +160,7 @@ class Cekit(object):
 
         return self
 
-    def run(self):
+    def configure(self):
         if self.args.verbose:
             logger.setLevel(logging.DEBUG)
         else:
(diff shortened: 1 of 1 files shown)


cekit/cli.py
```diff
@@ -171,31 +171,36 @@ def run(self):
...
+                                   self.args.target,
+                                   self.args.build_engine,
+                                   self.args.overrides,
+                                   params)
 
-            # We have all overrided repo fetch so we can discover modules
-            # and process its dependency trees
-            discover_modules(os.path.join(self.args.target, 'repo'))
+        # Now we can fetch repositories of modules (we have all overrides)
+        get_dependencies(self.generator.image,
```
dbecvarik: this is  cloning the module repos I would postpone that to run method.

jmtd:
The motivation for this PR is to make writing tests in #272 easier

cekit/cli.py
jmtd: Thanks, good point: force-pushed a rebase to fix.

dbecvarik:
thanks!

PR #282, opened 2018-08-06
Support defining help_template in image.yaml

jmtd:
I think this is likely to be the way this is most often used in
practice. An image author may wish to provide a custom help.md file,
potentially containing long-form static text, either directly in an
image.yaml, or indirectly via a module (perhaps to share with a
family of images).

<s>I've marked this WIP because I think the precedence should be

> defined in config < defined in image.yaml (post module processing) < defined on cmdline

But right now that's not how things are, config file trumps image.yaml,
and we haven't implemented command line argument for this at all.</s>
precedence fixed; let's not add a new cmdline argument, instead users could do `--overrides { help: template: foo.md }` or similar

### DIFF
--- a/cekit/descriptor/image.py
+++ b/cekit/descriptor/image.py
@@ -22,7 +22,8 @@ map:
   modules: {type: any}
   packages: {type: any}
   osbs: {type: any}
-  volumes: {type: any}""")
+  volumes: {type: any}
+  help_template: {type: text}""")
 
 
 def get_image_schema():
--- a/cekit/descriptor/image.py
+++ b/cekit/descriptor/image.py
@@ -23,7 +23,9 @@ map:
   packages: {type: any}
   osbs: {type: any}
   volumes: {type: any}
-  help_template: {type: text}""")
+  help:
+    map:
+      template: {type: text}""")
 
 
 def get_image_schema():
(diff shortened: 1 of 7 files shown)


cekit/descriptor/image.py
```diff
@@ -22,7 +22,8 @@
   modules: {type: any}
   packages: {type: any}
   osbs: {type: any}
-  volumes: {type: any}""")
+  volumes: {type: any}
+  help_template: {type: text}""")
```
dbecvarik: I'm not sure if we would like to have something like dictionary here so we can pass more parameteres in future..

jmtd: Ah perhaps a help: key and a dict underneath with template: and possibly other stuff in future?

docs/descriptor/image.rst
```diff
@@ -8,6 +8,7 @@ Image descriptor contains all information Cekit needs to build and test a contai
 .. include:: version.rst
 .. include:: description.rst
 .. include:: from.rst
+.. include:: help.rst
```
dbecvarik: please include this into overrides/module descriptor pages too...

dbecvarik:
Also can you please add some doc section into operation guide showing how to display resulting docs etc? so people are aware of how this work and can be checked?

docs/descriptor/image.rst
jmtd: Sorry can you give me an example of how you would like to see this incorporated?

dbecvarik:
thanks!

PR #284, opened 2018-08-08
generate io.openshift.expose-services label

jmtd:
As per the spec here:
https://docs.openshift.com/enterprise/3.2/creating_images/metadata.html

This will silence some dockerfile linter things we see.

Problems:

 - [x] documenting the new port keys
 - [x] <s>service is mandatory in the spec. should we make it mandatory in the YAML?</s> seems too proscriptive. Instead don't append ports w/o a service description, so as to not generate an invalid tag
 - [x] duplicating labels. We should generate only if there is no user-supplied label, and
       probably move the logic out of the template jinja and into the subroutine that generates
       default missing values before template processing
 - [x] tests

### DIFF
--- a/cekit/generator/base.py
+++ b/cekit/generator/base.py
@@ -2,6 +2,7 @@
 
 import logging
 import os
+import socket
 
 from jinja2 import Environment, FileSystemLoader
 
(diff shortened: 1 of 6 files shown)


cekit/generator/base.py
```diff
@@ -126,6 +127,32 @@ def override(self, overrides_path):
...
+                else:
+                    # attempt to supply a service name by looking up the socket number
+                    try:
+                        service = socket.getservbyport(p['value'], p.get('protocol','tcp'))
+                        r += ":{}".format(service)
+                        ports.append(r)
+
+                    except OSError: # py3
+                        pass
+                    except socket.error: # py2
```
jmtd: Having two `except` clauses is failing the  Codacy/PR Quality Review. In Python 2, we could write `except OSError, socket.error` but that is not valid in Python 3. The issue is two different exception names/classes are used in 2 vs 3.

PR #288, opened 2018-08-16
Implemented way to run particlar test scenario only

dbecvarik:
Cekit now have ``--test-name`` switch whcih can be used to run
specific scenario only


jmtd:
Looks good; but since test_tags is not processed at all if `--test-name` is specified, would it be worth making `--test-name` and the other command line argument mutually exclusive (using `.add_mutually_exclusive_group()`)? Not sure if the `--test-wip` argument should be handled similarly too

PR #292, opened 2018-08-23
Moved behave invocation from subprocess into API

dbecvarik:


goldmann:
Very nice

dbecvarik:
it worked for me.. I've tried --test-wip and --test-name too. So it should be ok.

goldmann:
Great, thanks!

PR #300, opened 2018-08-28
WIP: PoC runing tests in dist-git dir

dbecvarik:


cekit/builders/osbs.py
```diff
@@ -248,6 +251,8 @@ def add(self):
         for d in ["repos", "modules"]:
             # we probably do not care about non existing files and other errors here
             subprocess.call(["git", "add", "--all", d])
+        if os.path.exists("modules/tests"):
```
goldmann: This is probably not needed sine we already add whole modules dir above, right?

dbecvarik: no we are adding only used modules.. this is just PoC I've already refactored lot of the code...

PR #304, opened 2018-09-10
Implemented translation for ODCS ppc repositories

dbecvarik:


cekit/generator/osbs.py
```diff
@@ -2,13 +2,18 @@
...
 
+from cekit import tools
 from cekit.generator.base import Generator
 
 logger = logging.getLogger('cekit')
 
+PPC_REPOS_TABLE = {}
+PPC_REPOS_TABLE['rhel-7-server-rpms'] = 'rhel-7-for-power-le-rpms'
+PPC_REPOS_TABLE['rhel-7-extras-rpms'] = 'rhel-7-for-power-le-extras-rpms'
+PPC_REPOS_TABLE['rhel-server-rhscl-7-rpms'] = 'rhel-7-server-for-power-le-rhscl-rpms'
```
goldmann: I would see a general repos map:

```python
repos = {}
repos['rhel-7-server-rpms'] = {'x86_64': 'rhel-7-server-rpms', 'ppc64le': 'rhel-7-for-power-le-rpms'}
...
```

goldmann:
Additionally, repositories are listed twice for `x86_64` arch.

goldmann:
Generation of both, `content_sets.yml` and `container.yaml` should be done directly on dist-git sources. We could also generate these files as it is done now to get the idea what part of these files are generated. Other idea is to fetch the specific files at the generation time, copy them to the `target` directory, modify (or create) and copy them again to dist-git repository.

PR #307, opened 2018-09-10
Made overrides and overrides-file options not mutually exclusive

dbecvarik:


goldmann:
It would be great to have tests here:

```
cekit build --overrides-file a.yaml --overrides-file b.yaml --overrides '{"labels": {"test": "aaa"}}' 
```

Of course files `a.yaml` and `b.yaml` should also modify the selected label.

dbecvarik:
added tests

PR #311, opened 2018-09-12
WIP: Introduce "docgen" command: module doc generation

### LINKED ISSUE #268: Add feature to generate module documentation
Consider this python script in @rcernich's fork of cct_module (look a bit familiar?):

https://github.com/rcernich/cct_module/blob/CLOUD-2477/generate_docs.py

Should we incorporate that functionality into cekit?

It would be a good thing to have it maintained somewhere properly for other module authors to use. The idea is it will generate a markdown (or asciidoc) file that could then be committed alongside a module source code, so that there was a good API documentation for the module in the source, browseable on GitHub.

Cekit as a place for this code: on the one hand this is not how cekit works for its other features (input: image YAML, output: Dockerfile etc); on the other hand, much of the machinery needed to implement this well (parsing module yaml, jinja integration, schemas etc) is in cekit already, and putting this functionality in would mean it was a one-stop-shop for image-or-module authors.

Thoughts @goldmann @dbecvarik ?

jmtd:
This is an initial port of generate_docs.py from rcernich's fork
of cct_modules to a cekit command:

> https://github.com/rcernich/cct_module/blob/840739f604af2b3fdced61cf94d1626a4e059dbb/generate_docs.py

This command does not require an image descriptor (image.yaml) which
all the other commands do, therefore it hooks in at a different place
and as part of this initial implementation is effectively mutually
exclusive to those other commands. The UI around that needs some
tweaking.

Fixes  #268.

Todo (suggest more!)

 - [ ] fix UX around `cekit docgen` and/or interaction with other commands
 - [ ] docs
 - [x] tests
 - [ ] rebase and sign-off all commits


cekit/docgen.py
```diff
@@ -0,0 +1,61 @@
...
+        self.template = Template(open(self.template_file).read())
+        self.generate_doc_for_module(self.descriptor)
+
+    module_dirs = ['./jboss']
+
+    def generate_doc_for_module(self, module_file):
+        with open(module_file) as open_file:
+            data = yaml.load(open_file)
+        output_file = os.path.join(os.path.dirname(module_file), 'README.adoc')
+        print ("Generating %s..." % os.path.join(os.path.relpath(output_file, os.getcwd())), os.path.basename(output_file))
```
dbecvarik: this should be logger.info() probably

cekit/docgen.py
```diff
@@ -0,0 +1,61 @@
...
+"""
+
+class Docgen():
+    def __init__(self):
+        # find module descriptor
+        self.descriptor = 'module.yaml'
+        if not os.path.isfile(self.descriptor):
+            logger.error("can't find module.yaml descriptor")
+            sys.exit(1)
+        logger.debug("module descriptor path {} ".format(self.descriptor))
```
dbecvarik: Messages usually starts with uppercase letter

cekit/docgen.py
```diff
@@ -0,0 +1,61 @@
...
+            data = yaml.load(open_file)
+        output_file = os.path.join(os.path.dirname(module_file), 'README.adoc')
+        print ("Generating %s..." % os.path.join(os.path.relpath(output_file, os.getcwd())), os.path.basename(output_file))
+        with open(output_file, "w") as text_file:
+            text_file.write(autogen_warning)
+            text_file.write(self.template.render(data))
+
+    def scan_for_modules(self, curdir=os.getcwd()):
+        if not os.path.isdir(curdir):
+            return
```
dbecvarik: I would use sys.exit(0)

jmtd: I agree, although I think I might eventually rewrite this whole routine around `os.walk`

jmtd:
This seems to work for 106 of the 109 module.yaml in cct_module right now. The broken ones are

 - [ ] os-eap-logging (artifacts w/o name or path)
 - [ ] os-eap-activemq-rar  (same)
 - [ ] jboss.container.jokokia.bash (same)

jmtd:
all modules now generate (w/o throwing tracebacks) but the output is not necessarily great, yet

goldmann:
@jmtd I think it would be bad to loose this PR, are you willing to take a look at it again and prepare for review again?

jmtd:
Hey @goldmann , sure I'll try. It's a pretty ugly PR at the moment but I agree it would be good to have this feature. (I'd also like some debug verbs in cekit that would e.g. print out some representation of module dependencies from the Module objects that are built).

cekit/docgen.py
```diff
@@ -0,0 +1,59 @@
...
+import logging
+import os
+import yaml
+import sys
+from cekit.template_helper import TemplateHelper
+from jinja2 import Environment, Template, FileSystemLoader
+
+logger = logging.getLogger('cekit')
+
+autogen_warning="""////
```
jmtd: might as well move this out of code and into the template

cekit/docgen.py
jmtd: I might cut this entirely for this PR and possibly bring it back in a later one. It's still not hooked into the "docgen" verb anyway.

cekit/template_helper.py
```diff
@@ -50,3 +50,24 @@ def ports(self, available_ports):
                 port_list.append(p.get('value'))
 
         return port_list
+
+    def artifacts(self, inarts):
```
jmtd: this is only necessary because the "docgen" action is being handled before the module initialisation (in fact, the module initialisation never happens). If the module init had happened, the in-memory modules would have missing name fields calculated.

cekit/templates/template.adoc.jinja
```diff
@@ -0,0 +1,150 @@
+{%- macro resource_location(resource) -%}
```
jmtd: this filename (template.adoc.jinja) should maybe more descriptive (module.adoc.jinja?)

jmtd:
Current status/problems with output 

- [ ] Environment Variables" section is present but with empty tables (table present with header but no rows)
- [ ] "execute" key and scripts are not listed in the output
- [ ] list of "Included Modules" should probably be a (bullet) list
- [ ] version does not appear in output
- [ ] "Resources" section has empty table if no resources are present
- [ ] "artifacts" don't show up in the output

goldmann:
@jmtd Do we still want to carry it?

jmtd:
@goldmann hmm. I think there's still a need to document modules and a helper for that is very welcome. but we do have that other python script already. Shall we try to discuss it at the coordination meeting Tomorrow?

goldmann:
@jmtd Sure, let's do this!

goldmann:
Closing this PR, it's stalled.

PR #314, opened 2018-09-18
fixes parallel writes to artifact cache index

dbecvarik:


cekit/cache/artifact.py
```diff
@@ -37,43 +42,35 @@ def add(self, artifact):
...
+            cache_entry.update({alg: chksum})
 
-        self._update_cache(cache)
+        self._update_cache(cache_entry, artifact_id)
+        return artifact_id
 
     def delete(self, uuid):
         cache = self._get_cache()
         artifact = cache.pop(uuid)
         os.remove(os.path.expanduser(artifact['cached_path']))
```
goldmann: Why not just?

```
os.remove(os.path.join(self._cache_dir, uuid))
```

Tne you would not need to to even get it from cache above.

dbecvarik: I need to get it from cache above as if its not cached I'll print error

cekit/cache/artifact.py
```diff
@@ -9,26 +10,30 @@
...
 
     def _get_cache(self):
-        if not os.path.exists(self._cache_index_file):
-            return {}
-        with open(self._cache_index_file, 'r') as file_:
-            return yaml.safe_load(file_)
+        cache = {}
+        for index_file in glob.glob(os.path.join(self._cache_dir, '*.yaml')):
+            with open(index_file, 'r') as file_:
+                cache[index_file[len(self._cache_dir)+1:-5]] = yaml.safe_load(file_)
```
goldmann: os.path.basename(path)?

PR #317, opened 2018-09-25
Test are not executing full generate anymore.

dbecvarik:
The test command now is not generating Dockerfile. It also doesn't
fetch any artifacts and is not processing any package repositories.

### DIFF
--- a/cekit/cli.py
+++ b/cekit/cli.py
@@ -227,7 +227,6 @@ class Cekit(object):
             # if tags are not specified on command line we take them from image descriptor
             if not self.args.tags:
                 self.args.tags = generator.get_tags()
-            print(generator.image)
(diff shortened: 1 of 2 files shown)


cekit/cli.py
```diff
@@ -220,22 +220,23 @@ def run(self):
...
+            if self.args.build_tech_preview:
+                generator.generate_tech_preview()
 
-            # we run generate for every possible command
-            if self.args.commands:
-                generator.prepare_modules()
+            # if tags are not specified on command line we take them from image descriptor
+            if not self.args.tags:
+                self.args.tags = generator.get_tags()
+            print(generator.image)
```
goldmann: Remove it :)

PR #327, opened 2018-10-31
Added microdnf support

ruromero:
Signed-off-by: Ruben Romero Montes <rromerom@redhat.com>

Allow to choose between `yum` or `microdnf` package managers

Fix https://github.com/cekit/cekit/issues/328

### DIFF
--- a/cekit/templates/template.jinja
+++ b/cekit/templates/template.jinja
@@ -1,3 +1,29 @@
+{% macro repo_create(pkg_manager, repositories) -%}
+    {% if pkg_manager in ['microdnf', 'yum'] %}
+        {% for repo in repositories %}
+repos/{{ repo.filename }} \
+        {% endfor %}
+    /etc/yum.repos.d/
+    {% endif %}
+{%- endmacro %}
+{% macro repo_remove(pkg_manager, repositories) -%}
+    {% if pkg_manager in ['microdnf', 'yum'] %}
+rm{% for repo in repositories %} /etc/yum.repos.d/{{ repo.filename }}{% endfor %}
+    {% endif %}
+{%- endmacro %}
+{% macro repo_install(pkg_manager, rpm) -%}
+    {% if pkg_manager in ['microdnf', 'yum'] %}
+{{ pkg_manager }} install -y {{ rpm }} \
+    && {{ pkg_manager }} clean all && rm -rf /var/cache/yum
+    {% endif %}
+{%- endmacro %}
+{% macro pkg_install(pkg_manager, packages) -%}
+    {% if pkg_manager in ['microdnf', 'yum'] %}
+{{ pkg_manager }} install -y {%- for package in packages %} {{ package }}{% endfor %} \
+    && {{ pkg_manager }} clean all && rm -rf /var/cache/yum \
+    && rpm -q {% for package in packages %}{{ package }}{% endfor %}
+    {% endif %}
+{%- endmacro %}
 # Copyright 2017 Red Hat
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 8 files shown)


cekit/templates/template.jinja
```diff
@@ -32,16 +32,16 @@ COPY \ {% for repo in packages.repositories_injected %}
 {% if packages.repositories %}
 {% for repo in packages.repositories %}
 {% if repo.present and repo.rpm %}
-RUN yum install -y {{ repo.rpm }} \
-    && yum clean all && rm -rf /var/cache/yum
+RUN {{ pkg_manager }} install -y {{ repo.rpm }} \
```
rcernich: This seems a bit fragile, as it assumes the subcommands are the same between package managers.  Perhaps it might be better to stub this out into separate macros that are invoked based on the specified package manager, e.g. `{{ dnf_install(packages) }}' or `{{ yum_install(packages) }}`.

ruromero:
@rcernich can you review? I added both inside the same conditional because they are interchangeable. However this opens the possibility to provide more managers like apt-get and such.

ruromero:
@dbecvarik I added the autocomplete and docs, can you review the overall as well?
Thanks

dbecvarik:
Docs is ok, but there are now some tests failures.

dbecvarik:
Let's wait for @rcernich but for me its pretty ok.

cekit/templates/template.jinja
```diff
@@ -1,3 +1,29 @@
...
+{%- endmacro %}
+{% macro repo_install(pkg_manager, rpm) -%}
+    {% if pkg_manager in ['microdnf', 'yum'] %}
+{{ pkg_manager }} install -y {{ rpm }} \
+    && {{ pkg_manager }} clean all && rm -rf /var/cache/yum
+    {% endif %}
+{%- endmacro %}
+{% macro pkg_install(pkg_manager, packages) -%}
+    {% if pkg_manager in ['microdnf', 'yum'] %}
+{{ pkg_manager }} install -y {%- for package in packages %} {{ package }}{% endfor %} \
```
rcernich: I was thinking that you would have a separate macro for each package manager, e.g. `dnf_install` and `yum_install`.  That said, the rest of the changes are above and beyond.

ruromero: Why adding multiple macros when you can have one for all? I think it'd be easier to extend
I decided to add macros for:
* create/install/delete a repo
* install a package

cekit/templates/template.jinja
jmtd: This is true, but I'd argue not a blocker. The commands are common at least for yum, dnf and microdnf

cekit/templates/template.jinja
jmtd: Indeed, support for e.g. Debian is more than just changing to apt; the package names will be different, the paths will be different…

cekit/templates/template.jinja
```diff
@@ -23,30 +49,24 @@ USER root
...
+RUN {{ repo_install(pkg_manager, repo.rpm) }}
 
 {% endif %}
 {% endfor %}
 {% endif %}
 
 # Install required RPMs and ensure that the packages were installed
-RUN yum install -y {%- for package in packages.install %} {{ package }}{% endfor %} \
-    && yum clean all && rm -rf /var/cache/yum && \
-    rpm -q {% for package in packages.install %} {{ package }}{% endfor %}
```
jmtd: NACK at the moment because this line is not represented in the new macro. With `yum`, it will return a success status code if any package you asked for is installed: even if another did not. This line is necessary to ensure the build fails if a package fails to install. Please incorporate that into the macro.

cekit/templates/template.jinja
ruromero: The idea here (at least "mine") is to allow cekit to work with other images than centos/rhel. In case you use Debian, your image.yml will define different packages to install or else you will build it with an overrides file.

cekit/templates/template.jinja
ruromero: It is there https://github.com/cekit/cekit/blob/4a3e007e3aefbfd3c34b736ef0f01fc3c86b367a/cekit/templates/template.jinja#L22-L24

dbecvarik:
@jmtd can you re-review changes? I would like to merge it, before my other PRs so if there will be conflicts it will be me who will solve them :)

dbecvarik:
@ruromero  merged, thanks for contribution!

PR #329, opened 2018-11-07
apply modules as individual layers

rcernich:

### DIFF
--- a/cekit/builders/docker_builder.py
+++ b/cekit/builders/docker_builder.py
@@ -6,6 +6,7 @@ import yaml
 
 from cekit.builder import Builder
 from cekit.errors import CekitError
+from docker_squash.squash import Squash
 
 try:
     docker_client = docker.Client(version="1.22")
@@ -25,6 +26,7 @@ class DockerBuilder(Builder):
             params = {}
         self._tags = params.get('tags', [])
         self._pull = params.get('pull', False)
+        self._base = params.get('base')
         super(DockerBuilder, self).__init__(build_engine, target, params)
 
     def check_prerequisities(self):
@@ -62,18 +64,21 @@ class DockerBuilder(Builder):
                     line = yaml.safe_load(line)['errorDetail']['message']
                     raise CekitError("Image build failed: '%s'" % line)
 
-                if '---> Running in ' in line:
-                    docker_layer_ids.append(line.split(' ')[-1])
-
-                if '---> Using cache' in build_log[-1]:
-                    docker_layer_ids.append(line.split(' ')[-1])
-
                 if line != build_log[-1]:
                     # this prevents poluting cekit log with dowloading/extracting msgs
                     log_msg = ANSI_ESCAPE.sub('', line).strip()
                     for msg in log_msg.split('\n'):
                         logger.info('Docker: %s' % msg)
                     build_log.append(line)
+
+                    if '---> Running in ' in line:
+                        docker_layer_ids.append(line.split(' ')[-1].strip())
+                    elif 'Successfully built ' in line:
+                        docker_layer_ids.append(line.split(' ')[-1].strip())
+                    elif '---> Using cache' in build_log[-2]:
+                        docker_layer_ids.append(line.split(' ')[-1].strip())
+
+            self.squash_image(docker_layer_ids[-1])
 
             for tag in self._tags[1:]:
                 if ':' in tag:
@@ -96,3 +101,9 @@ class DockerBuilder(Builder):
                     "your image/module descriptor for proper repository " \
                     " definitions."
             raise CekitError(msg, ex)
+
+    def squash_image(self, layer_id):
+        logger.info("Squashing image %s..." % (layer_id))
+        # XXX: currently, cleanup throws a 409 error from the docker daemon.  this needs to be investigated in docker_squash
+        squash = Squash(docker=docker_client, log=logger, from_layer=self._base, image=layer_id, tag=self._tags[0], cleanup=False)
+        squash.run()
--- a/cekit/cli.py
+++ b/cekit/cli.py
@@ -233,7 +233,8 @@ class Cekit(object):
                           'pull': self.args.build_pull,
                           'redhat': config.get('common', 'redhat'),
                           'target': self.args.build_osbs_target,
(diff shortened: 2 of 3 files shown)


dbecvarik:
As we released Cekit 2.2. We should probably merge this to develop if this is the way to go so its in the beginning of the development cycles, so we have time to analyse impact.
@rcernich can you please rebase?

rcernich:
Rebased.  It appears to be working.

cekit/descriptor/image.py
```diff
@@ -45,48 +50,318 @@ def __init__(self, descriptor, artifact_dir):
...
+    def process_defaults(self):
+        """Prepares default values before rendering"""
+        if not self.run:
+            self.run = Run({})
+
+        # do we want to force a user?
+        if 'user' not in self.run:
+            self.run._descriptor['user'] = cekit.DEFAULT_USER
+
+    @property
```
dbecvarik: If we want to go for properties, we should do it for all the descriptor objects and remove collections.MutableMapping from object hierarchy and consider removing __getitem__, __delitem__ and other dunder methods so we have only one and consistent way of accessing self._dictionary. This PR opens a way of inconsistency and is error prone as we can have different logic for accessing inner dictionary.

cekit/descriptor/image.py
```diff
@@ -45,48 +50,318 @@ def __init__(self, descriptor, artifact_dir):
...
+    def volumes(self):
+        return self.get('volumes', [])
+
+    @property
+    def help(self):
+        return self.get('help', {})
+
+    def apply_image_overrides(self, overrides):
```
dbecvarik: why we have different logic for applying overrides? it seems like code duplication. Its even using merge() method from descriptor.

cekit/descriptor/image.py
```diff
@@ -45,48 +50,318 @@ def __init__(self, descriptor, artifact_dir):
...
+                    image_modules[name] = module
+            self.modules._descriptor['install'] = image_modules.values()
+
+            if override.run != None:
+                if self.run:
+                    self.run = override.run.merge(self.run)
+                else:
+                    self.run = override.run
+
+    def apply_module_overrides(self, module_registry):
```
dbecvarik: its merging logic duplicated again. If its need to be tweaked why not to override merge() method in Overrides/Modules descriptors?

cekit/descriptor/image.py
rcernich: This is because the logic is different.  `apply_image_overrides()` applies the overrides specified on the command line, while `apply_module_overrides()` applies overrides coming from the modules.

If you look at the logic for `apply_module_overrides()`, you will see that it overrides the artifact definitions and module versions in the modules, based on what was specified in the image and/or overrides.  It also collects any package repositories specified on the modules (which should be considered deprecated, as we should only be declaring repositories on the image/overrides).  Lastly, it collects `run` settings, which will be merged into the image after all the modules have been processed.

There are a couple of reasons for processing things in this fashion.

1. It allows module versions to be overridden without affecting the order in which the modules are applied.
2. It keeps the module and image objects intact.  The only changes are the overrides that are applied, e.g. module version changes and artifact definitions, and in the case of the image, name, from, labels, envs, and the union of all run declarations.
3. Notice the order of the merge operations is reversed.

HTH

cekit/descriptor/image.py
rcernich: See comment above.  No duplication, different behavior.  Then only thing changed in the modules are the artifact definitions.  The rest of this just collects the required modules, package repositories, and the result of all run specifications in the modules.

At the end of this processing, the image object has all of its fields updated from overrides, as well as the result of merging the `run` field from all overrides and modules, and a single list of modules to be installed.  The template then proceeds to apply each module, in order, to the image (i.e. we're not merging all the content from all modules into a single descriptor anymore).

cekit/descriptor/image.py
rcernich: I can look at doing this as a follow on effort, but it's a bit of work and most likely error prone.

dbecvarik: its error prone, but its better to refactor and break things now than later after more testing.

cekit/descriptor/image.py
dbecvarik: Ok, but can it be move to Overrides.merge() Module.merge() respectively so its collocated with the descriptor?

rcernich: It's not merging, so no.  That's the purpose of this change.  There is no need to merge modules into the image.  The only merging that is being done is overrides into image, artifacts (from image and overrides) into modules, and the run field.

This requires a bit of a change of thinking.  With this, the overrides only override what is already specified, e.g. it can no longer be used to add modules or artifacts to an image, but can override what is already specified.  The other fields in the overrides (e.g. labels, etc.) only apply to the image itself.  The only thing modules override in the image is the run field, as there can be only one run field for the image.  This allows modules to configure run.  Other than that, everything is self contained.

rcernich:
Found the issue with the tests.  Needed to convert dict.values() to an actual list, e.g. list(dict.values()).

cekit/descriptor/image.py
dbecvarik: ok. I'm changing review to OK, but I'll suggest to refactor code to new @properties like access as code should be consistent. @goldmann any opinions?

dbecvarik:
Its probably good time to move tests to python 3.6 or newer, what is in current fedora. or tweak how tests env is created

rcernich:
I think updating the tests is orthogonal to this effort.  Let's get this merged so it can begin fermenting.

dbecvarik:
yes, I agree that tests are separate issue. it was just mention, that this can cause future issues

rcernich:
I added squash functionality, as the extra USER, ENV and LABEL instructions really bloat the resulting images.

rcernich:
I think the sooner we get this merged, the better.  Let it fester a bit.

rcernich:
can we please get this merged before i have to rebase again?

rcernich:
You might also want to cherry-pick the squash commit into master.

goldmann:
The squash commit won't be cherry-picked. We only do this for bugfixes. It'll land in master once we release new version.

PR #338, opened 2018-12-06
Fix support in plain artifacts handling for OSBS

goldmann:
If the artifact is a plain artifact and it does not exists in Brew, OSBS builds is failing.

It was fixed by falling back into the regular procedure (check local, remote cache).
If this fails, the build fails. If it doesn't, OSBS will use a mix of artifacts:
fetched using fetch-artifacts-url.yaml and from lookaside cache. Approriate
warnings are added.


goldmann:
Merging to build the RPM to test it on CI.

PR #342, opened 2018-12-11
WIP: Adding commit option flag for osbs builds

kyguy:


goldmann:
Initial question before you move forward: why do you want to add switch to commit? I think it does not make sense to execute the build without syncing the source. Source will be generated either case, it can be different from what is already synced. If you execute a build without syncing, the contract would be broken.

I don't see the benefit of executing it via Cekit over executing `fedpkg`/`rhpkg` manually.

What's the use case here?

kyguy:
I'm working on a cekit wrapper for automating the build/commit process for multi-container products in osbs.  

1. The wrapper would traverse all image directories, generating Dockerfiles for every directory containing an image descriptor file using cekit, 
2. Then the wrapper would collect these generated files into a separate directory and commit/push all these generated source files (the Dockerfiles, content_sets.yml, etc. for every image in the product) to dist-git in one commit. 
3. Then, the wrapper would traverse all image directories again, building all the images in dist-git using cekit (cekit build --build-engine=osbs). 

The changes in this PR would allow the wrapper to build all the images using cekit without committing a separate commit to dist-git per each image.

However, now I am thinking, that if I get step 2 right, then when I do the cekit builds with the wrapper, cekit will have nothing to commit as the wrapper has already made the changes to all the respective image sources in dist-git. So I might not need these changes after all. What do you think?

goldmann:
My take on it would be probably following:

1. Wrapper executes Cekit's generate phase to generate required resources.
2. Wrapper should collect generated resources for every image.
3. Wrapper commits collection of resources to dist-git.
4. Wrapper executes builds for the whole collection.

I think steps 2-4 are fairly easy to implement. And step 1 is something like this:

```
$ cekit -v generate --redhat --build-engine osbs
```

kyguy:
Sounds like a solid plan to me. It looks like I should be able to implement these steps without any changes to cekit too so I am going to close this PR. Thanks for the guidance @goldmann , much appreciated!

Issue #343, opened 2018-12-12
Do not run yum commands if there are no packages to be installed


goldmann:
Even when there are no packages to install, Cekit now runs `yum makecache`. This breaks if there are no repositories available.

Additionally, we should not run `yum clean all` if there were no packages installed.

```
2018-12-12 15:32:09,447 cekit        INFO     Using Docker builder to build the image.
2018-12-12 15:32:09,447 cekit        DEBUG    Building image with tags: 'amq7/amq-online-1-mqtt-gateway:1.0', 'amq7/amq-online-1-mqtt-gateway:latest'
2018-12-12 15:32:09,448 cekit        INFO     Building container image...
2018-12-12 15:32:09,529 cekit        INFO     Docker: Step 1/32 : FROM jboss/openjdk18-rhel7:1.1
2018-12-12 15:32:09,530 cekit        INFO     Docker: ---> 42a04589ff05
2018-12-12 15:32:09,530 cekit        INFO     Docker: Step 2/32 : USER root
2018-12-12 15:32:09,530 cekit        INFO     Docker: ---> Using cache
2018-12-12 15:32:09,531 cekit        INFO     Docker: ---> d473180fc74c
2018-12-12 15:32:09,531 cekit        INFO     Docker: Step 3/32 : RUN yum makecache
2018-12-12 15:32:09,572 cekit        INFO     Docker: ---> Running in 3fe4652596ef
2018-12-12 15:32:09,933 cekit        INFO     Docker: Loaded plugins: ovl, product-id, search-disabled-repos, subscription-manager
2018-12-12 15:32:09,949 cekit        INFO     Docker: This system is not receiving updates. You can use subscription-manager on the host to register and assign subscriptions.
2018-12-12 15:32:10,007 cekit        INFO     Docker: There are no enabled repos.
2018-12-12 15:32:10,007 cekit        INFO     Docker:  Run "yum repolist all" to see the repos you have.
2018-12-12 15:32:10,007 cekit        INFO     Docker:  To enable Red Hat Subscription Management repositories:
2018-12-12 15:32:10,007 cekit        INFO     Docker:      subscription-manager repos --enable <repo>
2018-12-12 15:32:10,008 cekit        INFO     Docker:  To enable custom repositories:
2018-12-12 15:32:10,008 cekit        INFO     Docker:      yum-config-manager --enable <repo>
[...]

PR #345, opened 2018-12-12
issue #343 ensure repos exist before running yum commands

rcernich:


rcernich:
Patch for #343

Issue #355, opened 2019-01-10
Review dependency management


goldmann:
- [ ] Find what is provided via Python library and what are external dependencies
- [ ] Categorize dependencies by task
- [ ] Make sure the code is gracefully covering missing dependencies
- [ ] Update documentation

Everything needs to be documented clearly so that user knows what is installed by default and what needs to be still added.

Related: #285 and #286.

goldmann:
Investigation of external libraries and when these are used in Cekit.

External tool | When required | Comment
------------ | --------------------------- | --------------
`buildah` | `build` command and `--build-engine buildah` parameter is used |
`docker` | `build` command and `--build-engine docker` parameter is used |
`odcs-client` | `generate` command and `--build-engine buildah` or `--build-engine docker` parameters are used | Only useful when `--redhat` parameter is used, not tested without it
`fedpkg` | `build` command and `--build-engine osbs` parameter is used |
`rhpkg` | `build` command and `--build-engine osbs` and `--redhat` parameters are used |
`rhpkg-stage` | `build` command and `--build-engine osbs` and `--redhat` and `--stage` parameters are used |
`git` | Multiple places like fetching modules | This is actually a hard dependency, without it Cekit doesn't work
`brewkoji` | Artifact is defined as [plain artifact](https://cekit.readthedocs.io/en/latest/descriptor/image.html#plain) and `--redhat` and `build` command and `--build-engine buildah` or `--build-engine docker` parameters are used |

goldmann:
Anything that is not a Python library is impossible to require directly in Python (like `requirements.txt`). This is obvious, it better to say it :)

The only meta packaging we currently maintain are RPMs. Spec files are located in the https://github.com/cekit/packaging repository and builds are available in Copr: https://copr.fedorainfracloud.org/coprs/g/cekit/cekit/.

Since some of dependencies that cannot be pulled in Python directly are required to make Cekit useful, we need to make sure the user is notified about it. It should be done like this:

1. All external requirements should be checked for existence when they are required:
    * Core at startup
    * Others, when required switches are used
1. In spec files we provide we need to ensure that core libraries are installed.
1. Add suggestions to install required tools, but do not require it.

The last step is the one that bothers me the most. Better solution to the problem would be to externalize features into plugins, where each plugin could be packaged as separate RPM. This way, if someone is interested in building Docker images, the builder plugin would be installed.

goldmann:
[...]

PR #357, opened 2019-01-11
[WIP] Dependency handling

goldmann:
So far implemented checking for core dependencies as well as generator dependencies.

Code in modified files was reformatted.

Related #355


goldmann:
Ok, this should be good to go, currently writing tests.

PR #383, opened 2019-02-13
fix small typo in cli.py

vorburger:


goldmann:
I approved this, but it looks like it targets master. we do not add any commits to master, but instead we work on develop for the next Cekit version. Only bugfixes are backported to master.

goldmann:
I tried to change the base to develop, but it resulted in fairly big mess, so I reverted this. Can you please open this PR against develop?

vorburger:
> Can you please open this PR against develop?

done

goldmann:
Thanks!

PR #395, opened 2019-02-14
Handle pip or Fedora RPM install for sphinx-autobuild. Make explicit …

rnc:
…the dependent packages


rnc:
@goldmann I've removed the change to requirements which isn't necessary. This change means that if sphinx-autobuild was installed through either pip (e.g. in a virtualenv or globally), or via python2 RPM, or via python3 RPM they will all work.

PR #398, opened 2019-02-18
Add some details on development requirements

rnc:
First draft of some extra documentation - how to setup development environment.


docs/CONTRIBUTING.rst
```diff
@@ -1,5 +1,89 @@
+
+
+Contributing to the project
+===========================
+
+We strongly advise to use `Virtualenv <https://virtualenv.pypa.io/en/stable/>`_ to install Cekit. Please consult your package manager for the correct package name.
```
goldmann: I would probably add some "prerequisities" or similar section here that would describe steps required to start the development process.

docs/CONTRIBUTING.rst
```diff
@@ -1,5 +1,89 @@
...
+- python3-virtualenv
+
+For documentation, the following RPMs apply ( ``port_for==0.3.1`` must still be installed manually )
+
+- python3-argh
+- python3-livereload
+- python3-pathtools
+- python3-sphinx
+- python3-sphinx-autobuild
+- python3-watchdog
```
goldmann: I'm unsure if listing all the RPMs is a good idea. This can become very easily out of date. If you think it makes sense, we can keep it. Just a note. I would personally just stop with the note above the list that system packages can be used with `--system-site-packages` switch.

docs/CONTRIBUTING.rst
```diff
@@ -1,5 +1,89 @@
...
+It is possible to ask virtualenv to inherit pre-installed system packages thereby reducing the virtualenv to a delta between what is installed and what is required. This is achived by using the flag ``--system-site-packages`` (See `here <https://virtualenv.pypa.io/en/latest/userguide/#the-system-site-packages-option>`_ for further information).
+
+.. note::
+
+   Every time you want to use Cekit you must activate Cekit Python virtual environment by executing ``source ~/cekit/bin/activate``
+
+.. note::
+   For those using ZSH a useful addition is `Zsh-Autoswitch-VirtualEnv <https://github.com/MichaelAquilina/zsh-autoswitch-virtualenv>`_ the use of which avoids the labour of manually creating the virtualenv and activating it each time.
+
+Currently within Fedora 29 almost all the required packages are available as RPMs; non-python packages are:
```
goldmann: Maybe instead of listing these packages we could just rely on the runtime dependency mechanism to suggest them? Maybe tests could use same mechanism?

rnc:
@goldmann The RH maintainer for `python3-sphinx-autobuild` has promised to release updated versions with my PR for F29/F30 - this will change the ansible script and the docs.

docs/CONTRIBUTING.rst
rnc: Moved to ansible script

docs/CONTRIBUTING.rst
rnc: Rephased.

docs/CONTRIBUTING.rst
rnc: The cekit runtime rpm does/should provide the runtime deps ; currently there is nothing for the dev side - I have provided a set of ansible scripts instead.

docs/installation.rst
```diff
@@ -77,3 +77,5 @@ If you don't want to (or cannot) use Virtualenv, the best idea is to install Cek
 .. include:: dependencies.rst
 
 .. include:: upgrade.rst
+
+.. include:: CONTRIBUTING.rst
```
goldmann: I don't think this should be part of the installation document.

rnc: Perhaps move it to the top level then?

goldmann: Yes, this should be better.

rnc:
Re ansible ; I only provided those to remove the rpms from the docs ( as it wasn't ideal there I agree). I'm not keen on `hack/ansible` perhaps `support/...` to imply non required but supporting files ? (although potentially then the completion files could go under there)

goldmann:
For me `support/` is fine.

rnc:
@goldmann  I think completion (bash/zsh) should also be moved under support directory?

goldmann:
@rnc It can, yes, but we need to remember to update spec files for it.

Also: https://github.com/cekit/cekit/issues/423

Issue #387, opened 2019-02-14
Detect and handle docker daemon not running


rnc:
Currently I get this trace out: ; I think it should be handled better.
```
    2019-02-14 09:35:39,961 cekit        INFO     Preparing resource 'odcs-60967.repo'
    2019-02-14 09:35:40,289 cekit        INFO     Handling artifacts...
    2019-02-14 09:35:40,289 cekit        INFO     Preparing resource 'hawkular-javaagent'
    2019-02-14 09:35:40,313 cekit        INFO     Preparing resource 'jolokia-jvm-1.5.0.redhat-1-agent.jar'
    2019-02-14 09:35:40,315 cekit        INFO     Preparing resource 'jmx_prometheus_javaagent'
    2019-02-14 09:35:40,316 cekit        INFO     Rendering Dockerfile...
    2019-02-14 09:35:40,434 cekit        INFO     Using Docker builder to build the image.
    2019-02-14 09:35:40,434 cekit        INFO     Building container image...
    Traceback (most recent call last):
      File "/usr/lib/python3.7/site-packages/urllib3/connectionpool.py", line 600, in urlopen
        chunked=chunked)
      File "/usr/lib/python3.7/site-packages/urllib3/connectionpool.py", line 354, in _make_request
        conn.request(method, url, **httplib_request_kw)
      File "/usr/lib64/python3.7/http/client.py", line 1229, in request
        self._send_request(method, url, body, headers, encode_chunked)
      File "/usr/lib64/python3.7/http/client.py", line 1275, in _send_request
        self.endheaders(body, encode_chunked=encode_chunked)
      File "/usr/lib64/python3.7/http/client.py", line 1224, in endheaders
        self._send_output(message_body, encode_chunked=encode_chunked)
      File "/usr/lib64/python3.7/http/client.py", line 1016, in _send_output
        self.send(msg)
      File "/usr/lib64/python3.7/http/client.py", line 956, in send
        self.connect()
      File "/usr/lib/python3.7/site-packages/docker/transport/unixconn.py", line 42, in connect
        sock.connect(self.unix_socket)
    FileNotFoundError: [Errno 2] No such file or directory
     
    During handling of the above exception, another exception occurred:
     
    Traceback (most recent call last):
      File "/usr/lib/python3.7/site-packages/requests/adapters.py", line 449, in send
        timeout=timeout
      File "/usr/lib/python3.7/site-packages/urllib3/connectionpool.py", line 638, in urlopen
[...]

PR #399, opened 2019-02-18
#387 , #388 : Provide better error handling when docker isn't running…

rnc:
…/setup

### DIFF
--- a/cekit/builders/docker_builder.py
+++ b/cekit/builders/docker_builder.py
@@ -4,8 +4,8 @@ import re
 import yaml
 import sys
 import traceback
+import requests
 
-from requests.exceptions import ConnectionError
 from cekit.builder import Builder
 from cekit.errors import CekitError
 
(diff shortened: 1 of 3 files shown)


cekit/builders/docker_builder.py
```diff
@@ -131,6 +136,27 @@ def build(self, build_args=None):
...
+                          "See https://developer.fedoraproject.org/tools/docker/docker-installation.html and " \
+                          "http://www.projectatomic.io/blog/2015/08/why-we-dont-let-non-root-users-run-docker-in-centos-fedora-or-rhel"
+            elif 'FileNotFoundError' in exception_chain:
+                message = "Unable to contact docker daemon. Is it started?"
+            else:
+                message = "Unknown ConnectionError from docker ; is the daemon started and correctly setup?"
+
+            if sys.version_info.major == 3:
+                # Work-around for python 2 / 3 code - replicate exception(...) from None
+                cekit = CekitError(message, ex)
```
goldmann: I would probably rather not use `cekit` as the variable name here. Use `cekit_error` or anything else.

rnc:
@goldmann What about the Codacy review ... I could try changing it so that I don't import ConnectionError directly  ( but just requests ) ?

cekit/builders/docker_builder.py
rnc: Ok no problem; thanks!

PR #412, opened 2019-02-26
CLI rewrite

### LINKED ISSUE #364: Review CLI
We need to think if current CLI is how it is should be done, or maybe it is better to have multiple parsers having parameters for specific targets (generate, build, etc).

We currently use argparse, maybe it would be worth to check other framework?

To review:
* How Cekit is currently or should be used?
* How to make sure what switch is for what action?

Useful to read:

* https://codeburst.io/building-beautiful-command-line-interfaces-with-python-26c7e1bb54df
* https://www.davidfischer.name/2017/01/python-command-line-apps/
* https://opensource.com/article/18/5/3-python-command-line-tools
* https://realpython.com/comparing-python-command-line-parsing-libraries-argparse-docopt-click/

goldmann:
Fixes #364

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ docs/_build
 *.egg-info
 .vscode
 Pipfile.lock
+venv
--- a/cekit/builders/buildah.py
+++ b/cekit/builders/buildah.py
@@ -11,10 +11,10 @@ logger = logging.getLogger('cekit')
 class BuildahBuilder(Builder):
     """This class representes buildah builder in build-using-dockerfile mode."""
 
-    def __init__(self, build_engine, target, params={}):
-        self._tags = params.get('tags')
-        self._pull = params.get('pull', False)  # --pull-always
-        super(BuildahBuilder, self).__init__(build_engine, target, params)
+    def __init__(self, common_params, params):
+        #self._tags = params.get('tags')
+        # self._pull = params.get('pull', False)  # --pull-always
+        super(BuildahBuilder, self).__init__('buildah', common_params, params)
 
     @staticmethod
     def dependencies():
@@ -27,12 +27,12 @@ class BuildahBuilder(Builder):
 
         return deps
 
-    def build(self):
+    def run(self):
         """Build container image using buildah."""
-        tags = self._tags
+        tags = self.params.tags
         cmd = ["sudo", "buildah", "build-using-dockerfile"]
 
-        if self._pull:
+        if self.params.pull:
             cmd.append('--pull-always')
 
         # Custom tags for the container image
--- a/cekit/builders/buildah.py
+++ b/cekit/builders/buildah.py
@@ -5,15 +5,13 @@ import os
 from cekit.builder import Builder
 from cekit.errors import CekitError
 
-logger = logging.getLogger('cekit')
+LOGGER = logging.getLogger('cekit')
 
 
 class BuildahBuilder(Builder):
-    """This class representes buildah builder in build-using-dockerfile mode."""
+    """This class represents buildah builder in build-using-dockerfile mode."""
 
     def __init__(self, common_params, params):
-        #self._tags = params.get('tags')
-        # self._pull = params.get('pull', False)  # --pull-always
         super(BuildahBuilder, self).__init__('buildah', common_params, params)
 
     @staticmethod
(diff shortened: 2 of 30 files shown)


cekit/cli.py
```diff
@@ -1,295 +1,347 @@
...
-    def error(self, message):
-        self.print_help()
-        sys.stderr.write('\nError: %s\n' % message)
-        sys.exit(2)
+    # Remove the default 'cli' command
+    # Conditional here to make tests work better
+    if 'cli' in commands:
+        commands.remove('cli')
 
+    # return Map({'commands': commands, 'params': Map(params)})
```
rnc: Question - why does the cli command need to be removed? ( for the tests ) ?

goldmann: It does not need to be removed in fact, because we test for existence specific commands in the list and `cli` is not one of them.

The comment about the added conditional is that when executed as part of testusite the "default" `cli` command is not added there. I'm not exactly sure now what was the reason. Making it conditional was the easiest way to move forward.

Besides this, I'm still unsure if such parsing will be the final one. I'll give a thought once I get into shape the other CLI (`cekit-cache`).

.gitignore
```diff
@@ -18,3 +18,4 @@ docs/_build
 *.egg-info
 .vscode
 Pipfile.lock
+venv
```
rnc: Shouldn't this be .venv?

goldmann: I use the `venv` directory for my virtualenvs. I can change it though.

cekit/builders/buildah.py
```diff
@@ -11,10 +11,10 @@
 class BuildahBuilder(Builder):
     """This class representes buildah builder in build-using-dockerfile mode."""
 
-    def __init__(self, build_engine, target, params={}):
-        self._tags = params.get('tags')
-        self._pull = params.get('pull', False)  # --pull-always
-        super(BuildahBuilder, self).__init__(build_engine, target, params)
+    def __init__(self, common_params, params):
+        #self._tags = params.get('tags')
+        # self._pull = params.get('pull', False)  # --pull-always
```
rnc: Shouldn't these two commented out lines be removed?

cekit/builders/osbs.py
```diff
@@ -27,40 +27,31 @@
...
+        self._rhpkg_set_url_repos = []
 
-        if params.get('redhat'):
-            if params.get('stage'):
-                self._fedpkg = 'rhpkg-stage'
-                self._koji = 'brew-stage'
+        if config.get('common', 'redhat'):
+            if self.params.get('stage'):
+                self._fedpkg = '/usr/bin/rhpkg-stage'
+                self._koji = '/usr/bin/brew-stage'
```
rnc: Question - isn't /usr/bin always in the path ? Why does it need to be explicitly specified?

goldmann: Its safer to specify absolute path to the binary. In other cases it may be a security issue. Yes we need to do it in several other places.

Pylint complains about this.

cekit/cli.py
```diff
@@ -1,313 +1,345 @@
...
+from cekit.tools import Map
 from cekit.version import version
 
-# FIXME we shoudl try to move this to json
+# FIXME we should try to move this to json
 setup_logging()
-logger = logging.getLogger('cekit')
-config = Config()
+LOGGER = logging.getLogger('cekit')
+CONFIG = Config()
```
rnc: Is there a reason why in some code its 'logger' and some its 'LOGGER' ? Could we have it all lower case?

goldmann: It should be upper case everywhere by Python convention. I haven't updated it everywhere yet.

goldmann: https://www.python.org/dev/peps/pep-0008/#constants

rnc: What are you doing replying at this time :-) I'e just realised its PEP8 having looked it up , sorry for the noise ;-)
Yeah, probably needs to be updated everywhere

.gitignore
rnc: Ah! Apologies, its because I was adding .venv (for https://github.com/MichaelAquilina/zsh-autoswitch-virtualenv/) and thought you were doing similar. No worries.

cekit/cli.py
goldmann: I could ask you the same thing :P

.gitignore
goldmann: I can move to this convention. Seems to be cleaner to have it in a  "hidden" directory.

cekit/builders/docker_builder.py
```diff
@@ -152,38 +142,51 @@ def _build_with_docker(self, docker_client):
             if "To enable Red Hat Subscription Management repositories:" in ' '.join(build_log) and \
                     not os.path.exists(os.path.join(self.target, 'image', 'repos')):
                 msg = "Image build failed with a yum error and you don't " \
-                      "have any yum repository configured, please check " \
+                      "have any yum repository qured, please check " \
```
rnc: Spelling?

goldmann: Yah, files besides `cli.py` were not entirely scanned yet for issues with spelling tools and pylint.

cekit/builders/osbs.py
rnc: Ok.

rnc:
Really nice restructure , impressive piece of work :+1:

goldmann:
Thanks!

PR #416, opened 2019-03-04
Properly define what artficats should be copied to dist-git

### LINKED ISSUE #414: OSBS builder doesn't work well with plain artifacts with target set
It uses the `name` attribute instead of `target` if available.

goldmann:
The issue was that if an artifact had the 'target' kwy set,
the OSBS builder incorrectly skipped it and tried to use 'name'.

Fixes #414


rnc:
:+1:

PR #417, opened 2019-03-04
Do not download artifacts from unsuccessful builds

### LINKED ISSUE #415: Handle deleted artifacts in koji
When an artifact is deleted we should not try to fetch it. Instead the build should be terminated.

goldmann:
We should fetch artifacts only when the build is in COMPLETE state.
In all other cases, build is in incorrect state, we should not fetch
any artifacts in such case.

Fixes #415

### DIFF
--- a/cekit/tools.py
+++ b/cekit/tools.py
@@ -83,6 +83,20 @@ def get_brew_url(md5):
         get_build_cmd = ['brew', 'call', '--json-output', 'getBuild', 'buildInfo=%s' % build_id]
         logger.debug("Executing '%s'" % " ".join(get_build_cmd))
         build = yaml.safe_load(subprocess.check_output(get_build_cmd))
+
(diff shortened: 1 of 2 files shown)


cekit/tools.py
```diff
@@ -83,6 +83,18 @@ def get_brew_url(md5):
...
+        # State 1 means: COMPLETE which is the only success state. Other states are:
+        #
+        # 'BUILDING': 0
+        # 'COMPLETE': 1
+        # 'DELETED': 2
+        # 'FAILED': 3
+        # 'CANCELED': 4
+        if build['state'] != 1:
+            raise CekitError(
+                "Artifact with checksum {} was found in Koji metadata but the build was removed making the artifact not available for downloading anymore".format(md5))
```
rnc: Being pedantic surely if state is 0 the build isn't removed, just not complete yet ? :-)

goldmann: Oh, yeah, good catch!

PR #424, opened 2019-03-07
Cache command improvements

### LINKED ISSUE #361: Review caching
- [x] `cekit-cache` command without parameters prints nothing
- [x] Caching local file should not use involve usage of cacher, even if defined
- [x] Error handling - in case an error occurs while caching artifact, leftovers should be cleaned up.
- [x] Artifact added to cache should have computed checksum for every supported algorithm
  * This also helps with the artifact duplication if the `add` command was run on with the same artifact, but with different algorithms.

goldmann:
Fixes #361


goldmann:
@rnc This is now ready for review. There are a few commits. You can review it all in once, but for better understating on the change I suggest to look at commit separately.

I suggest also to look at tasks listed in #361.

PR #428, opened 2019-03-11
Make shell completion files generic

### LINKED ISSUE #423: Update bash and zsh completions for new CLI
Related to #412.

goldmann:
With this the autocompletion feature for Click apps is enabled:
https://click.palletsprojects.com/en/7.x/bashcomplete/

These snippets were generated using following commands:

_CEKIT_COMPLETE=source cekit
_CEKIT_CACHE_COMPLETE=source cekit-cache

_CEKIT_COMPLETE=source_zsh cekit
_CEKIT_CACHE_COMPLETE=source_zsh cekit-cache

Fixes #423

### DIFF
--- a/support/completion/bash/cekit
+++ b/support/completion/bash/cekit
@@ -1,90 +1,43 @@
-_cekit_global_options()
-{
-    local options
-    options+='--tags '
-    options+='--overrides '
-    options+='--overrides-file '
-    options+='--target '
-    options+='--descriptor '
-    options+='--work-dir '
-    echo "$options"
+_cekit_completion() {
+    local IFS=$'
+'
+    COMPREPLY=( $( env COMP_WORDS="${COMP_WORDS[*]}" \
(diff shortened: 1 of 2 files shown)


rnc:
Completion using click looks nice.

I think it would be useful to put a paragraph in the guide (possibly installation section) regarding the completion. If I recall correctly the RPM installs it globally? But it would still need to be activated?

Also, if these are generated files, would it be useful to (possibly) mark that somehow and also provide a target within the makefile to generate/update them?

goldmann:
When the RPM with completion is installed, then it is available immediately (well, in the new shell you launch). If you want I can try to update the installation instructions to incorporate this, but I already have a PR in that area which will most probably be clashing: https://github.com/cekit/cekit/pull/433

rnc:
If its available upon installation implicitly then its probably not worthwhile.

PR #431, opened 2019-03-12
Move support for defining package managers to image definition

### LINKED ISSUE #400: package manager should possibly be selected via image.yaml / module yaml
Thanks to #328 we have `--package-manager`, but since the correct package manager is a property of the image being built, rather than the builder environment (running cekit), it should probably be something that can be put in a module or `image.yaml` property.

goldmann:
This removes support for defining package manager used to install packages
from the command line (and configuration file) to the image descriptor
itself.

Adds support for dnf.

Fixes #400

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -13,7 +13,7 @@ pytest-mock = "*"
 docker = "*"
 docker-squash = "*"
 click = "*"
-pylint = "<2.0.0"
+pylint = "*"
 autopep8 = "*"
 
 [packages]
--- a/Pipfile
+++ b/Pipfile
@@ -13,6 +13,8 @@ pytest-mock = "*"
 docker = "*"
 docker-squash = "*"
 click = "*"
+pylint = "<2.0.0"
+autopep8 = "*"
 
 [packages]
 pykwalify = ">=1.6.0"
--- a/docs/descriptor/packages.rst
+++ b/docs/descriptor/packages.rst
@@ -2,9 +2,10 @@ Packages
 --------
 
 To install additional RPM packages you can use the ``packages``
-section where you specify package names and repositories to be used.
+section where you specify package names and repositories to be used, as well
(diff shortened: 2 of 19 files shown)


goldmann:
I wonder if it would not make more sense to not specify the default at all and require the `manager` key when defining packages to install.

goldmann:
On the other hand, having `yum` as the default value work in all Fedora/RHEL cases...

luck3y:
> On the other hand, having `yum` as the default value works in all Fedora/RHEL cases...

+1, as its also the current default, it will just continue to work as expected.

Pipfile
```diff
@@ -13,6 +13,8 @@ pytest-mock = "*"
 docker = "*"
 docker-squash = "*"
 click = "*"
+pylint = "<2.0.0"
```
rnc: Why < version 2 ? It seems F29 has 2.1.1 ...

rnc:
Should this have documentation updates as well?

goldmann:
Yes, I'll update this PR.

Pipfile
goldmann: I missed it, it was automatically added by IDE.

goldmann:
@rnc Docs added. I additionally worked to make it build without any warnings.

docs/descriptor/packages.rst
```diff
@@ -16,12 +17,30 @@ section where you specify package names and repositories to be used.
...
 
-.. _repo:
+Package manager
+^^^^^^^^^^^^^^^^^^
+
+It is possible to define package manager used in the image
+used to install packages as part of the build process.
+
+.. note::
+    If you do not specify this key the default value is ``yum``.
```
rnc: Should this list possible options? e.g.
```suggestion
    If you do not specify this key the default value is ``yum``. Options are ``yum``, ``dnf``, and ``microdnf``
```

PR #442, opened 2019-03-15
Make Behave tester work again

goldmann:
The issue was that the search for the behave module was shaded by the `behave.py` file. Added basic test to make sure we can execute tests.


goldmann:
@rnc  I'm trying to debug why the test is failing on CI, I'll update the PR and let you know.

PR #473, opened 2019-04-11
update contributing.md

elmiko:
The link inside this document went to a 404 page, this change updates
the link to the current location for the guide.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,5 +1,5 @@
 Please take a look at our contribution guide. It may help you create high-quality content.
 
-https://docs.cekit.io/en/latest/contribution-guide/
+https://cekit.readthedocs.io/en/develop/contribution-guide/index.html
 
 Thank you!


goldmann:
This is a tricky one :) The 404 page is actually expected. When we merge it to master at the release time, it will be built on readthedocs and available under that link. I haven't found a better way to do it. Ideas?

And yes, we're in the docs-writing phase before the 3.0 release. There are still a few TODO's as you noticed.

elmiko:
ah, ok. sorry. should i just close this then?

goldmann:
Unless you think there is a better way how to manage it :)

elmiko:
no, i didn't necessarily have a better suggestion. it was just something i noticed as i started reading through the code to figure out how i might propose a patch. i wasn't aware that there was a process underway to fix this for the next release.

goldmann:
Yah, the thing is that the current development version of CEKit (default branch: `develop`) did a fairly big rewrite on docs. Once we release 3.0, this will be a correct, stable link for contribution details. I guess we need to live with it a week or two more.

But thanks for pointing it out!

Issue #471, opened 2019-04-10
Add sha512 to the list of accepted hash algorithms


elmiko:
**Describe the solution you'd like**
I would like to see a `sha512` field added to the [image artifact descriptors](https://cekit.readthedocs.io/en/develop/descriptor/image.html#artifacts). The value for this field would be a sha512 sum, all lowercase, with no spaces or linebreaks.

**Additional context**
I am currently encountering a situation where an upstream project (Apache Spark) that distributes only pgp signatures and sha512 sums with current releases. Our team is using cekit to build our images and we would like to continue to support this security feature without needing to manually calculate hash sums in a different manner than the upstream.

elmiko:
i browsed the code in this repository and saw the references about the various hash algorithms, but i don't see any implementation for the actual checksum checks. i don't know too much about how cekit works, but i gather there might be a backend that is wrapped by this project, i understand that sha512 might not be supported.

if i have missed something in the code, my apologies.

goldmann:
Thanks for the ticket!

We reuse `hashlib` module to provide this and it looks like `sha512` is supported there, so I do not see any reason why we couldn't add support for it.

```
$ python
Python 2.7.15 (default, Oct 15 2018, 15:26:09) 
[GCC 8.2.1 20180801 (Red Hat 8.2.1-2)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import hashlib
>>> hashlib.algorithms
('md5', 'sha1', 'sha224', 'sha256', 'sha384', 'sha512')
>>> 
```


Here is the relevant code in CEKit:

https://github.com/cekit/cekit/blob/45cab4657e3c925762dc33299ee5268b45b4e26e/cekit/crypto.py#L1-L17

elmiko:
haha, i should have looked harder. i guess i didn't realize you were using the names from that list as keys to hashlib.

i'll take a look at working up a pr =)

goldmann:
@elmiko Let me know if you need any help with it. We are going to cut a release very soon (next week maybe), it would be good to have this PR ready sooner than later :)

elmiko:
will do @goldmann !

goldmann:
Fixed in https://github.com/cekit/cekit/pull/475.

PR #475, opened 2019-04-12
add sha512 as an optional checksum

elmiko:
This change brings in the `sha512` as a possible key for checksum values
associated with the `artifact` directive. It also updates the relevent
tests and documentation.

closes issue #471

### DIFF
--- a/cekit/cache/cli.py
+++ b/cekit/cache/cli.py
@@ -38,7 +38,7 @@ def ls():
 @click.option('--sha256', metavar="CHECKSUM", help="The sha256 checksum of the artifact.")
 @click.option('--sha512', metavar="CHECKSUM", help="The sha512 checksum of the artifact.")
 def add(location, md5, sha1, sha256, sha512):  # pylint: disable=unused-argument
-    if not (md5 or sha1 or sha256, sha512):
+    if not (md5 or sha1 or sha256 or sha512):
         raise click.UsageError("At least one checksum must be provided")
 
     CacheCli.prepare().add(location, md5, sha1, sha256, sha512)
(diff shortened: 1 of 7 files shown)


tests/test_validate_cache.py
```diff
@@ -119,9 +120,9 @@ def test_cekit_cannot_add_artifact_without_checksum(tmpdir):
     result = run_cekit_cache(['--work-dir',
                               work_dir,
                               'add',
-                              artifact], 2)
+                              artifact], 1)
 
-    assert "At least one checksum must be provided" in result.output
+    assert "Cannot cache artifact without checksum" in result.output
```
elmiko: changing this test was the only thing that confused me. it seemed to be working before, but after my changes it broke and i'm not sure why. fixing it was just a matter of changing the expected result code and the error text.

cekit/cache/cli.py
```diff
@@ -36,11 +36,12 @@ def ls():
 @click.option('--md5', metavar="CHECKSUM", help="The md5 checksum of the artifact.")
 @click.option('--sha1', metavar="CHECKSUM", help="The sha1 checksum of the artifact.")
 @click.option('--sha256', metavar="CHECKSUM", help="The sha256 checksum of the artifact.")
-def add(location, md5, sha1, sha256):  # pylint: disable=unused-argument
-    if not (md5 or sha1 or sha256):
+@click.option('--sha512', metavar="CHECKSUM", help="The sha512 checksum of the artifact.")
+def add(location, md5, sha1, sha256, sha512):  # pylint: disable=unused-argument
+    if not (md5 or sha1 or sha256, sha512):
```
goldmann: Here is the issue, there should be no comma, but `or` string.

tests/test_validate_cache.py
goldmann: Added a comment above that points where the issue is.

cekit/cache/cli.py
elmiko: oops, moving too fast. good spot!

elmiko:
ok, added the `or` and reverted the other test.

thanks again for the help @goldmann =)

goldmann:
This is why we have reviews :)

goldmann:
@elmiko Thank you!

elmiko:
glad to help!

PR #480, opened 2019-04-16
If we install pkgs in discrete steps now, we probably need to clean the pkg cache every time

luck3y:


luck3y:
@goldmann Note this allows my microdnf build to complete, but the image is quite large at the end (>1gb).

So this isn't the whole story here.

goldmann:
We do this on purpose.

We create cache at the beginning of the flow and clear the cache at the end to make the build faster. In other case, the repository cache will be rebuilt on every package manager execution which will significantly slower the execution.

I don't think we gain much more by clearing the cache after every execution.

goldmann:
@errantepiphany @jmtd any opinions here?

luck3y:
@goldmann Based on email discussions, we can probably close this in favor of a pkg manager / OS fix I think.

goldmann:
Sure, I'm happy with this.

errantepiphany:
@goldmann No strong opinion. Will default to your guys' decision, thanks. :)

Issue #483, opened 2019-04-16
Handle UBI built-in repositories


goldmann:
**Describe the solution you'd like**
Currently the UBI images ship with a set of enabled repositories, but these repositories do not exist yet. Refreshing the metadata fails with an error preventing the image to build.

jmtd:
Maybe the content_sets handling code should delete all existing repo files? The warning it prints "The image has ContentSets repositories specified, all other repositories are removed!" is a little confusing already, if you don't know that when it's talking about repositories it doesn't mean yum repo files in the image, but rather other possible repositories stuff in the image yaml

goldmann:
This is not affecting CEKit only, internal systems, such as OSBS are having same issue.

The proposed solution in OSBS is to rewrite the .repo file from the image to point to internal Pulp repos. After GA these repositories will became available and there will be no issue.

So, to reiterate - there is a problem between now and GA date for local builds. For internal OSBS builds this will not be an issue. After GA this issue will not exist for local builds either.

jmtd:
> So, to reiterate - there is a problem between now and GA date for local builds. For internal OSBS builds this will not be an issue. After GA this issue will not exist for local builds either.

The wider issue — base image has repos defined that should be ignored due to content sets, but aren't — exists for other contexts, too, though, e.g. RHEL image built on top of Fedora builder:

```
STEP 4: RUN yum makecache
Loaded plugins: ovl, product-id, search-disabled-repos, subscription-manager
This system is not registered with an entitlement server. You can use subscription-manager to register.
There are no enabled repos.
 Run "yum repolist all" to see the repos you have.
 To enable Red Hat Subscription Management repositories:
     subscription-manager repos --enable <repo>
 To enable custom repositories:
     yum-config-manager --enable <repo>
Error: error building at step {Env:[PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin container=oci] Command:run Args:[yum makecache] Flags:[] Attrs:map[] Message:RUN yum makecache Original:RUN yum makecache}: error while running runtime: exit status 1
```

goldmann:
This should be closed now since UBI is GA.

PR #503, opened 2019-05-01
backup/restore yum.repos.d when using ContentSets

jmtd:
I don't think you will want this — we discussed the idea in #483 but I don't think you liked it. I'm just raising this FYI as I'm most likely going to use this to build some images now where the base image is shipped with a repo file that is invalid today but will work in the future, as a short term workaround.

---

If we are using ContentSets, back up /etc/yum.repos.d and delete
the contents prior to copying in the ODCS repo file. In clean-up,
delete yum.repos.d again and move the backup back into place.

### DIFF
--- a/cekit/templates/template.jinja
+++ b/cekit/templates/template.jinja
@@ -105,6 +105,10 @@ FROM {{ from }}
 USER root
 {% if packages.repositories_injected %}
 
+# back up existing repo files and clear them out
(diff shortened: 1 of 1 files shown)


goldmann:
@jmtd Since #505 is now merged, do you think this addresses this PR too? Such modifications would be to add to some module now, if needed. Now it should be possible.

PR #505, opened 2019-05-01
Do not invoke pkg_manager makecache

### LINKED ISSUE #501: yum makecache
Currently, in a base image, if the default .repo file in /etc/yum.repos.d/ contains invalid content (yes, I know it shouldn't) the initial `yum makecache` will fail the image build. The problem occurs before any user scripting can take place (rm'ing the invalid repo file, for example), so its a bit of a problem in some cases.

An easy, but perhaps problematic solution, is to try and run it as `yum makecache || true` and continue regardless, thereby allowing the injected odcs repos (if present) and subsequent user scripting to do the right thing later. If no injected repos are present the build will fail at the package installation phase anyway.

This is more of an item for discussion rather than a concrete suggestion for a fix, but there are situations where something like this could be useful.

cc: @jmtd

jmtd:
Fixes #501.


goldmann:
Thanks for this!

Issue #407, opened 2019-02-21
Review templates


goldmann:
Current template is very complex, we need to make it much simpler, especially when we will suggest people using their own templates. We need to setup a contract. What variables are available and what you can do with it.

- [ ] Review of current state
- [ ] Custom templates? How/where to define them?
  - [ ] Documentation of the language used
  - [ ] Documentation of available objects

goldmann:
Related #405.

goldmann:
Additionally, we should investigate if/how we want to make it possible to extend/modify templates.

Comment from @jmtd taken from #377:

> we should support a command-line substitution of the help template, in addition to config file based. This is so we could (more easily) generate say, some docs that can be more easily consumed by an upstream docs team, instead of the default template which is designed to end up being piped to man(1) and so lacks more useful mark up like tables

This one was about the documentation template, but the same applies actually to the main Dockerfile template and should be looked together.

goldmann:
We do not allow for custom templates anymore.

Issue #405, opened 2019-02-21
Remove licenses from generated Dockerfiles


goldmann:
We should investigate Dockerfiles headers for licenses. Possibly to make it possible to generate Dockerfiles without any license or select one. Currently the Apache license is hardcoded.

goldmann:
License text will be removed.

PR #512, opened 2019-05-13
Bump copyright year in the template

jmtd:
I wonder, it's a little weird putting RH in the default template, I suppose a future feature might be some way of extracting that and permitting people to inject their own. But so long as its there, let's bump the year


goldmann:
Related: https://github.com/cekit/cekit/issues/405

goldmann:
And this too: https://github.com/cekit/cekit/issues/407

goldmann:
Thanks!

PR #519, opened 2019-05-24
Allow for supplementary files to be added to dist-git repository

### LINKED ISSUE #394: Add support for supplementary dist-git items
We have a use case for committing another yaml file for CVP (gating.yaml) which would currently be ignored by cekit, is it possible to add capabilities within cekit to define a list of supplementary items that would be synced to the dist-git at runtime? Or at the very least, support this additional file like with content_sets.yaml.

Thank you!

goldmann:
If a directory called `dist` is found next to the image descriptor,
contents of this directory will be copied to the **root** of the
dist-git directory.

Fixes #394.

### DIFF
--- a/cekit/builders/osbs.py
+++ b/cekit/builders/osbs.py
@@ -19,7 +19,7 @@ from cekit.config import Config
 from cekit.builder import Builder
 from cekit.descriptor.resource import _PlainResource
 from cekit.errors import CekitError
-from cekit.tools import Chdir
+from cekit.tools import Chdir, copy_recursively
 
 LOGGER = logging.getLogger('cekit')
 CONFIG = Config()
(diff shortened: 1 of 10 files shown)


luck3y:
Not a big fan of the dist name, I think of that being used to build a pkg distribution - how about buildcfg, distcfg or builddesc / distcfg?

It would be nice if it were configurable in image.yaml :)

goldmann:
@rnc Do you have any comments?

goldmann:
Points for discussion are available here: https://github.com/cekit/cekit/issues/394#issuecomment-495525352

goldmann:
Some changes were made to the implementation.

* Default directory name is now `osbs_extra`
* The directory name can be redefined within the image descriptor with the `extra_dir` key in the `osbs` section:
    ```yaml
    osbs:
      extra_dir: custom_name
    ```

Any comments?

AntStephenson:
Looks good to me :) Thanks!

goldmann:
Because this implementation got positive feedback, I'm going to merge it now. Thanks!

PR #523, opened 2019-05-27
Add a line break to large labels

### LINKED ISSUE #506: Labels with a very large value break container builds
**Describe the bug**
If the image.yaml descriptor has a very huge label value it will make cekit build fails (tested with docker and osbs and podman).

Apparently, the label string value are being parsed to bytes somewhere and this issue happens: 
https://pastebin.com/fG2EdUKP

I tried to do this, but the build failed with "Docker unknown instruction":

`if line != build_log[-1]:
    if isinstance(line, bytes): 
        line = line.decode('utf-8') `

With podman, this issue happens:

> Error: error building at step {Env:[USER_NAME=apb USER_UID=1001 BASE_DIR=/opt/apb HOME=/opt/apb PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin container=oci JBOSS_IMAGE_NAME=rhpam-7/rhpam-apb JBOSS_IMAGE_VERSION=1.1] Command:label Args:[] Flags:[] Attrs:map[] Message:LABEL  Original:LABEL }: LABEL requires at least one argument

**To reproduce**
use a image.yaml file descriptor with this label:
https://pastebin.com/CbbxgMGm

**Logs**
Paste any logs you have, use CEKit verbose output: `cekit -v`.
https://pastebin.com/fG2EdUKP

**Environment information:**
 - Operating system: Fedora 29
 - CEKit version: 3.0.1

**Additional context**
This label is a requirement to work with APB (ansible playbook bundle)

ricardozanini:
I was unsure about where to add the unit test, please fell free to ask to move/change this one to some other place. The `yaml` and `Image` imports seems odd in this test case.

Fixes #506

Signed-off-by: Ricardo Zanini <zanini@redhat.com>

### DIFF
--- a/cekit/generator/base.py
+++ b/cekit/generator/base.py
@@ -2,6 +2,7 @@
 
 import logging
 import os
+import re
 import platform
 import subprocess
 import shutil
@@ -90,6 +91,11 @@ class Generator(object):
         # we will persist cekit version in a label here, so we know which version of cekit
         # was used to build the image
         image_labels.append(Label({'name': 'io.cekit.version', 'value': cekit_version}))
+
+        for label in image_labels:
+            if len(label.value) > 128:
+                # breaks the line each time it reaches 128 characters
+                label.value = "\\\n".join(re.findall("(?s).{,128}", label.value))[:]
 
         # If we define the label in the image descriptor
         # we should *not* override it with value from
--- a/tests/test_unit_generator_docker.py
+++ b/tests/test_unit_generator_docker.py
@@ -7,10 +7,12 @@ import os
 from contextlib import contextmanager
 
 import pytest
+import yaml
 
 from cekit.config import Config
 from cekit.errors import CekitError
 from cekit.generator.docker import DockerGenerator
+from cekit.descriptor import Image
 
 
 odcs_fake_resp = b"""Result:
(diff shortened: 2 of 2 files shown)


cekit/generator/base.py
```diff
@@ -91,6 +92,11 @@ def add_build_labels(self):
         # was used to build the image
         image_labels.append(Label({'name': 'io.cekit.version', 'value': cekit_version}))
 
+        for label in image_labels:
+            if len(label.value) > 128:
+                # break the first line
```
goldmann: This comment is not entirely true, since it's breaking the line multiple times, if the line is multiple times longer than 128 characters.

tests/test_unit_generator_docker.py
```diff
@@ -86,6 +88,24 @@ def test_prepare_content_sets_should_not_fail_when_cs_is_empty(tmpdir):
...
+      - name: 'the.large.label'
+        value: Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec pretium finibus lorem vitae pellentesque. Maecenas tincidunt amet.
+    """), 'foo')
+    with docker_generator(tmpdir) as generator:
+        generator.image = image
+        with cekit_config(redhat=True):
+            generator.add_build_labels()
+            for label in generator.image.labels:
+                if label.name == 'the.large.label':
+                    assert "\\\n" in label.value
```
goldmann: Why not assert the the entire value here?

tests/test_unit_generator_docker.py
```diff
@@ -86,6 +88,24 @@ def test_prepare_content_sets_should_not_fail_when_cs_is_empty(tmpdir):
...
+    labels:
+      - name: 'the.large.label'
+        value: Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec pretium finibus lorem vitae pellentesque. Maecenas tincidunt amet.
+    """), 'foo')
+    with docker_generator(tmpdir) as generator:
+        generator.image = image
+        with cekit_config(redhat=True):
+            generator.add_build_labels()
+            for label in generator.image.labels:
+                if label.name == 'the.large.label':
```
goldmann: This is also something that is probably not needed. How about referencing the single label as:

```python
generator.image.labels[0]
```

ricardozanini: The `add_build_labels` function adds more labels to the list, that's why I added the loop instead of looking into the position. If the function changes in the future, might breck the test case. But I agree that is cleaner referencing the index instead. wdyt? Keep the loop?

cekit/generator/base.py
ricardozanini: wdyt of `breaks the line each time it reaches 128 characters`?

tests/test_unit_generator_docker.py
ricardozanini: Well, looking again, the function is appending values to the list. Unlikely that this will break the test in the future. Changing.

goldmann: Even if it will break in the future - it's fine. This means that the contract was broken and it's good to know about it.

goldmann:
Thank you!

PR #555, opened 2019-06-27
Add support for validating image sources

### LINKED ISSUE #551: Add a new option to only validate the image and module descriptors
Add a new option to validate only the image and module descriptors to find, for example, syntax errors, invalid values, etc.

Today the --dry-run option will perform all steps that the conventional build commands does, but not build the image.

**Additional context**
A validate option would be very useful to use in a CI environment where er can't use Red Hat internal resources.

goldmann:
This adds a --validate switch to the build command. It interrupts the build process
early in the stage, just after the model is prepared and validated. This
is similar to --dry-run with the difference that --dry-run additionally generates
required files for the selected builder (which --validate does not).

This required redesigning build hooks which are now following:

1. prepare()
2. before_generate()
3. generate()
4. before_build()
5. run()

Fixes #551

### DIFF
--- a/cekit/builder.py
+++ b/cekit/builder.py
@@ -29,15 +29,10 @@ class Command(object):
     def prepare(self):
         pass
 
-    def before_run(self):
-        pass
-
     def run(self):
         raise CekitError(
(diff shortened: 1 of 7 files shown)


spolti:
Very nice, thanks @goldmann

PR #560, opened 2019-07-03
[KEYCLOAK-10694] Add workaround for microdnf RH BZ#1700341 issue

iankko:
Per https://github.com/rpm-software-management/microdnf/issues/27 libdnf seems to have an issue on cache data, when doing multiple / subsequent package installs

Therefore clean the microdnf cache after the each install command
to avoid this

Also add a test to check the 'clean all' snippet is present in resulting Dockerfile, when using microdnf as a package manager

Signed-off-by: Jan Lieskovsky <jlieskov@redhat.com>

### DIFF
--- a/cekit/templates/template.jinja
+++ b/cekit/templates/template.jinja
@@ -19,6 +19,9 @@ rm{% for repo in repositories %} /etc/yum.repos.d/{{ repo.filename }}{% endfor %
 {% macro pkg_install(pkg_manager, packages) -%}
     {% if pkg_manager in helper.SUPPORTED_PACKAGE_MANAGERS %}
 {{ pkg_manager }} --setopt=tsflags=nodocs install -y {%- for package in packages %} {{ package }}{% endfor %} \
+    {% if pkg_manager == 'microdnf' %}
+    && {{ pkg_manager }} clean all \
+    {% endif %}
     && rpm -q{% for package in packages %} {{ package }}{% endfor %}
     {% endif %}
 {%- endmacro %}
--- a/tests/test_validate.py
+++ b/tests/test_validate.py
@@ -462,6 +462,40 @@ def test_override_add_module_and_packages_in_overrides(tmpdir):
     assert check_dockerfile_text(image_dir, 'COPY \\\n    test \\\n    /tmp/artifacts/')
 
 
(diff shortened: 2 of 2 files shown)


iankko:
@goldmann @jmtd PTAL once got a chance

Thanks!, Jan

goldmann:
Thanks for this! I think we can add this until a proper fix arrives. Can you add a test that would check if the `clean all` command was applied? I suggest to add it to `test_validate.py`. You will see there examples of tests that check Dockerfile generated content.

iankko:
Re: add test, on it, stay tuned.

In between,

```
# grep manager image.yaml 
      manager: microdnf
```

without this change CEKit 3.1.+ with `microdnf` running multiple pkg installs ends up with error like the following one:

```
...
2019-07-03 12:33:56,973 cekit        INFO     Docker: 
2019-07-03 12:33:57,920 cekit        INFO     Docker: Downloading metadata...
2019-07-03 12:34:00,759 cekit        INFO     Docker: error: cannot move /var/cache/yum/metadata/ubi-8-baseos-8-x86_64/packages to /var/cache/yum/metadata/ubi-8-baseos-8-x86_64.tmp/packages
2019-07-03 12:34:00,937 cekit        ERROR    You can look inside the failed image by running 'docker run --rm -ti 158cb8ffcc50 bash'
2019-07-03 12:34:00,937 cekit        ERROR    Image build failed, see logs above.
```

With this change being applied,  image build works:
```
...
2019-07-03 13:24:38,018 cekit        INFO     Squashing finished!
2019-07-03 13:24:45,630 cekit        INFO     New squashed image ID is 86560704ab9ca86b4eaf001b0e0040e135afae7bd9ec065520421bd330e97b31
2019-07-03 13:26:17,566 cekit        INFO     Done
2019-07-03 13:26:17,610 cekit        INFO     Image built and available under following tags: redhat-sso-cd-tech-preview/sso-cd-openshift:6, redhat-sso-cd-tech-preview/sso-cd-openshift:latest
2019-07-03 13:26:17,611 cekit        INFO     Finished!
```

iankko:
@goldmann All done, (Working) test added too

goldmann:
Thanks!

PR #561, opened 2019-07-03
Copy modules and artifacts when module is processed

### LINKED ISSUE #544: Reuse Docker layers
I would like to only rebuild starting from the module I modified. As an example:
https://github.com/wildfly/wildfly-s2i/blob/master/wildfly-builder-image/image.yaml

If I have:
from X
install
 module1
 module2
 module3
Any change to module3 should only rebuild the image starting from module3.

Today we are rebuilding all.

goldmann:
This chnges the way we generate Dockerfile so that
artifacts and modules are not copied at the beginning of
the Dockerfile. This makes it possible to reuse
Docker (and other builder's) cache for content that
was not changed.

To make use of rapid development with cached you need to
make sure the module you modify is installed later in the process
because this invalidates cache and all modules installed
later will need to be executed again.

If you use Docker builder, use the `--no-squash` option.
Without this, the squash post processing will remove all
intermediate container images created at the original image build time
removing the cache at the same time.

Fixes #544

### DIFF
--- a/cekit/descriptor/image.py
+++ b/cekit/descriptor/image.py
@@ -276,9 +276,8 @@ class Image(Descriptor):
                 artifact_overrides[name] = artifact
                 # add it to the list of everything
                 self._all_artifacts[name] = artifact
-                if name in image_artifacts:
-                    # apply override to image descriptor
-                    image_artifacts[name] = artifact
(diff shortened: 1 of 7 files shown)


goldmann:
Thanks!

PR #562, opened 2019-07-03
Fix for ZSH completion

### LINKED ISSUE #554: zsh completion does not work
**Describe the bug**
- cekit command completion in the zsh shell does not work. After typing "cekit " and then a tab, instead of the list of possible cekit command options being presented, one gets the current list of sub directories.

**To reproduce**
1. `sudo dnf install cekit cekit-bash-completion cekit-zsh-completion`
2. `zsh (if not already in zsh shell)`
3. `cekit <space> <tab>`

**Expected behavior**
- A list of possible cekit command options should appear.

**Logs**
- N/A

**Environment information:**
- OS: Fedora release 30 (Thirty) x86_64
- Kernel: 5.1.12-300.fc30.x86_64
- Shell: zsh 5.7.1
- CEKit version: 3.1.0

**Additional context**
- bash completion works as expected.

rnc:
Fixes #554

### DIFF
--- a/support/completion/zsh/_cekit
+++ b/support/completion/zsh/_cekit
@@ -1,7 +1,11 @@
+#compdef cekit cekit-cache
+
 _cekit_completion() {
     local -a completions
     local -a completions_with_descriptions
     local -a response
+    (( ! $+commands[cekit] )) && return 1
+
     response=("${(@f)$( env COMP_WORDS="${words[*]}" \
                         COMP_CWORD=$((CURRENT-1)) \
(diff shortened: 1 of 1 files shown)


rnc:
Note : you will need to remove any and all `.zcompdump*` files then rerun compinit to ensure this is picked up.

goldmann:
Maybe it would also make sense to send a PR upstream? https://github.com/pallets/click/blob/master/click/_bashcomplete.py#L41

PR #564, opened 2019-07-04
Deprecate OSBS --tech-preview switch

### LINKED ISSUE #559: Deprecate OSBS --tech-preview switch
Replacement for #257.

Related: https://github.com/jboss-openshift/cct_module/pull/263#issuecomment-400208056

Tests and images should not use `-tech-preview` suffix. Instead it should be added only to released image.

goldmann:
Additionally added documentation about this switch.

Fixes #559


rnc:
While the comment on https://github.com/cekit/cekit/issues/559 made sense while you are deprecating it do you think the deprecation note needs a reason/explanation in the docs - and what the replacement is?

goldmann:
This is a valid point, I'll clarify it in the docs.

goldmann:
@rnc This is now updated.

PR #565, opened 2019-07-04
Extend local module development documentation

### LINKED ISSUE #527: Extend local develoment guidelines to cover including changes in modules
Currently we do not have any information what would be the best practice to modify a module and what would be expected to be rebuilt locally after this change. Sometimes the image needs to be rebuilt entirely and this takes time. We need to explain it and suggest modules orders. It is hard to do since every image is different, but we should be able to put together some general guidelines.

goldmann:
* Added information about the name key
* Moved note about module install and caching
* Other small formatting improvements

Fixes #527

### DIFF
--- a/docs/guidelines/local-development.rst
+++ b/docs/guidelines/local-development.rst
@@ -1,13 +1,18 @@
 Local development
 ==========================
 
+.. contents::
+    :backlinks: none
+
 Developing image locally is an important part of the workflow. It needs to provide
 a simple way to reference parts of the image we changed. Executing a local build with our
 changes should be easily done too.
 
+Module development
+--------------------------------
 
 Referencing customized modules
---------------------------------
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 CEKit enables you to use a work in progress modules to build the image by using
 its overrides system. As an example, imagine we have very simple image which is using
@@ -27,6 +32,7 @@ one module from a remote Git repository, like this:
         # Add a shared module repository located on GitHub. This repository
         # can contain several modules.
         - git:
+            name: common
             url: https://github.com/cekit/example-common-module.git
             ref: master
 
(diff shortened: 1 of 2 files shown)


docs/guidelines/local-development.rst
```diff
@@ -42,27 +48,68 @@ repository locally by executing:
...
+      - name: common
+        path: /home/user/repo/cct_module
 
-Now we can build the image using overridden module by executing:
+Now we can build the image with Docker using overridden module by executing:
 
 .. code-block:: bash
 
-  $ cekit generate --overrides-file overrides.yaml
+  $ cekit build --overrides-file overrides.yaml docker
```
rnc: Spelling of overrides is different to above

docs/guidelines/local-development.rst
```diff
@@ -42,27 +48,68 @@ repository locally by executing:
...
+^^^^^^
+
+Below you can see suggestions that should make developing modules easier.
+
+Always define name for module repositories
+*******************************************
+
+We use the ``name`` key as the resource identifier in all places. If you do not define the ``name``
+key yourself, we will generate one for you. This may be handy, but in cases where you plan to use
+overrides it may be much better idea to define them.
```
rnc: Question - is the generated name unique?

docs/guidelines/local-development.rst
goldmann: Nice catch!

docs/guidelines/local-development.rst
goldmann: It is not really a unique name. It's based on the resource provider data. In case of git repositories it is the name of the repo IIRC, bit I would need to double-check in the code.

goldmann: I was right: https://github.com/cekit/cekit/blob/f75f098a74be1b50ec58fcbb195a49a0222ddcde/cekit/descriptor/resource.py#L288-L290

So, for `https://github.com/cekit/cekit.git` it will result in `cekit.git` as the `name` key.

goldmann:
Typo fixed, ready for review.

PR #566, opened 2019-07-05
Image help pages rewrite

### LINKED ISSUE #528: Image help pages should include environment variables defined in modules
The help generation should take into account the env var defined by the modules. Currently it is only generating doc for env var defined in image.yaml file.

goldmann:
Generated help pages now contain more output which is now even better
structured which should make it more readable.

Additionally, all labels and environment defined in modules
are now preseted in the help page.

Fixes #528


goldmann:
Example help page generated: https://paste.fedoraproject.org/paste/EYhBIKqQgtnVWmbFhR4WWQ.

PR #568, opened 2019-07-08
Clarify module versioning and enforce requirement for it

### LINKED ISSUE #194: Handling multiple module repositories containing modules with the same name
This might relate to #160 

If you include multiple module repositories in an image YAML, and reference a module name that occurs in both repositories, which repo wins? it's not clear that this is a defined situation and you will get the same behaviour every time.

### LINKED ISSUE #196: Cekit lacks the information about module versioning
Cekit has very basic support for versioning modules and requesting precise module version. We should work on this story end finish to address issues like #194 

Very brief info is here https://concreate.readthedocs.io/en/latest/descriptor.html#module-installation . but it needs to be expanded/reworked/tested and properly documented

goldmann:
This change enforces the requirement that is already in the documentation
that a module should have a version defined.

Code related to module versioning handling was rewritten.

Documentation was created to cover module versioning, both
from the handbook pov as well as from best practices view.

Fixes #196
Fixes #194


goldmann:
Thank you!

PR #569, opened 2019-07-09
Add warning when repository name is not specified.

### LINKED ISSUE #198: Add warning/error when repository name is not explicitly specified
The absence of a repository name makes it difficult to use overrides for building from local source.  I think there is another issue to document the auto-naming, which is mostly a problem for git repos, since the name includes the branch and .git suffix.  I'm guessing this is so you could use multiple versions of a repo in the same build, but is this something that would actually be done and if so, why?

rnc:
Fixes #198

### DIFF
--- a/cekit/descriptor/execute.py
+++ b/cekit/descriptor/execute.py
@@ -27,8 +27,8 @@ class Execute(Descriptor):
         descriptor['module_name'] = module_name
 
         if 'name' not in descriptor:
-            logger.warning("No value found for 'name'; using auto-generated value of %s/%s" %
-                           (module_name, descriptor['script']))
+            logger.warning("No value found for 'name'; using auto-generated value of {}/{}".
+                           format(module_name, descriptor['script']))
             descriptor['name'] = "%s/%s" % (module_name,
                                             descriptor['script'])
 
--- a/cekit/descriptor/execute.py
+++ b/cekit/descriptor/execute.py
@@ -1,3 +1,4 @@
+import logging
 import yaml
 
 import cekit
@@ -13,6 +14,8 @@ container_schemas = [yaml.safe_load("""
         seq:
           - {type: any}""")]
 
+logger = logging.getLogger('cekit')
+
 
 class Execute(Descriptor):
     def __init__(self, descriptor, module_name):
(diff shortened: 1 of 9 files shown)


cekit/descriptor/resource.py
```diff
@@ -80,6 +80,12 @@ def __ne__(self, other):
             return not self['name'] == other['name']
         return NotImplemented
 
+    @staticmethod
+    def empty_descriptor(default_value, container=""):
```
goldmann: By looking at the name of this method is not easy to get what it actually does.

How about creating a property instead? It could use an abstract method to generate default values which would be implemented in the other XYZResource classes. Or something similar? WDYT?

cekit/descriptor/resource.py
```diff
@@ -226,15 +233,16 @@ def _download_file(self, url, destination, use_cache=True):
 
 class _PathResource(Resource):
 
-    def __init__(self, descriptor, directory, **kwargs):
-        # if the path si relative its considered relative to the directory parameter
-        # it defualts to CWD, but should be set for a descriptor dir if used for artifacts
+    def __init__(self, descriptor, directory, container="", **kwargs):
```
goldmann: I would probably not add another parameter and use instead the "resource" phrase. It is used only in the log message, right?

cekit/descriptor/resource.py
```diff
@@ -226,15 +233,16 @@ def _download_file(self, url, destination, use_cache=True):
 
 class _PathResource(Resource):
 
-    def __init__(self, descriptor, directory, **kwargs):
-        # if the path si relative its considered relative to the directory parameter
+    def __init__(self, descriptor, directory, container, **kwargs):
```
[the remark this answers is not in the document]
rnc: Actually yes, I'll check but I think its only from `artifact` sections without `name`.

cekit/descriptor/resource.py
rnc: Thinking on this, its kind of pointless, its wrapping a single log line. Eliminating it might be more logical

cekit/descriptor/resource.py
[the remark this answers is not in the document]
rnc: @goldmann  What configures the 'git' resource?

cekit/descriptor/execute.py
```diff
@@ -24,6 +27,8 @@ def __init__(self, descriptor, module_name):
         descriptor['module_name'] = module_name
 
         if 'name' not in descriptor:
+            logger.warning("No value found for 'name'; using auto-generated value of %s/%s" %
```
goldmann: Generally we should use new style formatting with `.format()`

PR #579, opened 2019-07-15
When certifi is available, add a warning

### LINKED ISSUE #577: Try to detect certifi package installs and warn user if custom installation is found
The certifi library when installed locally provides a custom certificate store that can override the globally provided setting. We should try to find if certifi is installed and if yes, we should warn the user that certificate-related issues may appear.

We probably should do this only when `--redhat` switch is used.

goldmann:
This helps to debug issues related to certificate validation
issues where the certifi library overrides what is provided
system-wide.

Fixes #577


goldmann:
Thanks!

PR #584, opened 2019-07-18
Improve logging - show less warnings

goldmann:
This shows less warning, especially for modules without version specified. Without this change it would be printed 3 times. Additionally changes logging level to debug for missing name in scripts. We do not want to spam users with it.

### DIFF
--- a/cekit/descriptor/execute.py
+++ b/cekit/descriptor/execute.py
@@ -23,13 +23,14 @@ class Execute(Descriptor):
         super(Execute, self).__init__(descriptor)
 
         descriptor['directory'] = module_name
-
         descriptor['module_name'] = module_name
 
         if 'name' not in descriptor:
-            logger.warning("No value found for 'name'; using auto-generated value of {}/{}".
-                           format(module_name, descriptor['script']))
-            descriptor['name'] = "%s/%s" % (module_name, descriptor['script'])
+            # Generated name
+            descriptor['name'] = "{}/{}".format(module_name, descriptor['script'])
+
+            logger.debug("No value found for 'name' key in the execute section of the '{}' module; using auto-generated value: '{}'".format(
+                module_name, descriptor['name']))
 
     @property
     def name(self):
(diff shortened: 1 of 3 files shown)


cekit/descriptor/execute.py
```diff
@@ -23,13 +23,14 @@ def __init__(self, descriptor, module_name):
...
         descriptor['module_name'] = module_name
 
         if 'name' not in descriptor:
-            logger.warning("No value found for 'name'; using auto-generated value of {}/{}".
-                           format(module_name, descriptor['script']))
-            descriptor['name'] = "%s/%s" % (module_name, descriptor['script'])
+            # Generated name
+            descriptor['name'] = "{}/{}".format(module_name, descriptor['script'])
+
+            logger.debug("No value found for 'name' key in the execute section of the '{}' module; using auto-generated value: '{}'".format(
```
rnc: Why debug? I thought we were trying to warn people?

cekit/generator/base.py
```diff
@@ -182,8 +182,9 @@ def copy_modules(self):
         """
         target = os.path.join(self.target, 'image', 'modules')
         for module in self.image.modules.install:
-            module = self._module_registry.get_module(module.name, module.version)
-            LOGGER.debug("Copying module '{}' required by '{}'.".format(module.name, self.image.name))
+            module = self._module_registry.get_module(module.name, module.version, surpress_warnings=True)
```
rnc: surpress -> suppress

cekit/descriptor/execute.py
goldmann: Yes, but not necessarily for names in the scripts, which is **very** unlikely to be overriden. Nobody is defining a name for it, I would call it internal.

cekit/generator/base.py
goldmann: Fixed!

cekit/descriptor/execute.py
goldmann: In other words - printing a warning for the missing `name` key in the execute section of the module doesn't make sense. We can print it, but as a debug. Main use case is that modules are overridden as a whole and it does not make much sense to override particular execute scripts.

All modules define install scripts without defining the name, depending on the image complexity we would end up with many warnings, where fixing it would not give us anything because of the reason above.

PR #610, opened 2019-09-04
Show schema for artifacts in documentation

### LINKED ISSUE #607: Show schema for artifacts in documentation
It would be good to try to show the complete schema for, at least, artifacts in documentation. This could be extended on all the other descriptor elements.

goldmann:
This is a PoC where the schema of particular element is shown
in the documentation. This supports only artifacts for now,
but it could be possibly extended for other elements
in the future.

Fixes #607

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -77,7 +77,7 @@ jobs:
             mkdir -p ./venv
             virtualenv ./venv
             . venv/bin/activate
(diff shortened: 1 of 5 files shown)


goldmann:
Schemas can be seen here: https://1608-102005331-gh.circle-artifacts.com/0/home/circleci/project/docs/_build/html/descriptor/image.html#artifact-types

Issue #593, opened 2019-08-20
cekit-3.3.1 creates more layers than 3.2.0 and fails docker build locally


luck3y:
Dockerfiles attached below. cekit-3.3.1 exceeds the docker max limit which looks to be around 224 or so, which is about 50% more than the corresponding build with 3.2.0 which in the last build I counted about 160 or so layers before squash.

I haven't looked at doing any optimization of the image yet, as we really just needed to get them built, but we can take a look at that.

Image.yaml: https://github.com/jboss-container-images/jboss-eap-7-openshift-image/blob/eap73-openjdk11-dev/image.yaml

Dockerfiles:

[Dockerfile-3.3.1.txt](https://github.com/cekit/cekit/files/3522865/Dockerfile-3.3.1.txt)
[Dockerfile-3.2.0.txt](https://github.com/cekit/cekit/files/3522867/Dockerfile-3.2.0.txt)

FYI @goldmann @spolti

luck3y:
Note with 3.3.1 the total number of steps reported by cekit is 213. With 3.2.0 this is 153.

The failure: 
`2019-08-20 17:44:37,225 cekit        INFO     Docker: Step 211/213 : USER 185
2019-08-20 17:44:37,412 cekit        INFO     Docker: ---> Running in cd8329cd7838
2019-08-20 17:44:37,467 cekit        ERROR    You can look inside the failed image by running 'docker run --rm -ti 119b9e61215f bash'
2019-08-20 17:44:37,467 cekit        ERROR    ('Image build failed, see logs above.', CekitError("Image build failed: 'max depth exceeded'"))
Traceback (most recent call last):
  File "/usr/lib/python3.7/site-packages/cekit/builders/docker_builder.py", line 89, in _build_with_docker
    raise CekitError("Image build failed: '{}'".format(error_message))
cekit.errors.CekitError: Image build failed: 'max depth exceeded'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python3.7/site-packages/cekit/cli.py", line 334, in run
    command.execute()
  File "/usr/lib/python3.7/site-packages/cekit/builder.py", line 66, in execute
    self.run()
  File "/usr/lib/python3.7/site-packages/cekit/builders/docker_builder.py", line 242, in run
    image_id = self._build_with_docker(docker_client)
  File "/usr/lib/python3.7/site-packages/cekit/builders/docker_builder.py", line 150, in _build_with_docker
[...]

PR #613, opened 2019-09-11
Skip copying content for modules that have no execute

luck3y:
See https://github.com/cekit/cekit/issues/593

### DIFF
--- a/cekit/templates/template.jinja
+++ b/cekit/templates/template.jinja
@@ -92,9 +92,11 @@ rm{% for repo in repositories %} /etc/yum.repos.d/{{ repo.filename }}{% endfor %
         {% endfor %}
         {% endif -%}
 
+        {% if module.execute %}
         {% if not is_image %}
         # Copy '{{ module.name }}' {{ module_type }} content
         COPY modules/{{ module.name }} /tmp/scripts/{{ module.name }}
+        {% endif -%}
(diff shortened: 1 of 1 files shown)


luck3y:
@goldmann I'm going to add some info to the issue for discussion on this.

luck3y:
I'm taking a look at the tests as well.

tests/test_validate.py
```diff
@@ -609,7 +609,7 @@ def test_run_path_artifact_brew(tmpdir, caplog):
 
     img_desc = image_descriptor.copy()
     img_desc['artifacts'] = [{'name': 'aaa',
-                              'md5': 'd41d8cd98f00b204e9800998ecf8427e',
```
luck3y: Note this is a legit md5 in brew, so the fetch actually works, I changed it to something that doesn't exist (for now), but it might be better to pick something like: deadbeefdeadbeefdeadbeefdeadbeefdea not that that wouldn't occur either I suppose, but the intent is a bit clearer.

luck3y:
@goldmann I updated the tests to reflect the change, but I assume you may have a visceral reaction to this change. I definitely think that its a bug to add the copy in the case of no content, however.

goldmann:
FYI: I created a tracker issue: #614.

Fixes #614.

PR #625, opened 2019-10-02
Add rh-requirements.txt: for RH workflows

jmtd:
These packages are optional/soft Cekit dependencies that are required
for some of the workflows most likely used by Red Hat engineers.

This is a time-saver for those who install cekit via pip into a virtualenv. We set up an entirely new virtualenv for each cekit release, and need to then install these same dependencies each time.

### DIFF
--- a/rh-requirements.txt
+++ b/rh-requirements.txt
@@ -1,0 +1,16 @@
+# Here are listed optional Cekit dependencies that are needed for
+# specific Cekit workflows used in Red Hat.
+
+# for docker build back-end
+docker       >=4.0.2
+docker-squash>=1.0.7
+
+# for content sets
+odcs         >=0.2.35
+
+# for unit tests
+behave       >=1.2.6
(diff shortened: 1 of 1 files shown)


goldmann:
Wouldn't it better to be hosted by specific teams in their own repositories? There are different requirements (although some may be common, yes) and workflows.

jmtd:
> Wouldn't it better to be hosted by specific teams in their own repositories? There are different requirements (although some may be common, yes) and workflows.

It's not relating to an image source so I don't think it should live in an image source repository. I thought the advice for RH image maintainers was use cekit via virtualenv, but if that's not the case, perhaps a straw poll of who would find this useful?

goldmann:
Oh, sorry, I was thinking not about the image repository, but a repository that is shared across team members. I'm pretty sure every team has something like this. If a particular team finds it useful, it could be added to such repo.

Issue #631, opened 2019-10-31
osbs configuration is overridden when using --redhat


hmlnarik:
When initiating a build that uses `--redhat` defaults and overrides image.yaml `osbs` entry with a YAML passed in via one of the `--override` options, the `osbs` entry is ignored.

**To reproduce**
Have `my-overrides.yaml` file that contains `osbs` / `configuration` / `container` entry. Then run:

    cekit --redhat build--overrides-file my-overrides.yaml osbs

**Expected behavior**
`container.yaml` is generated.

**Logs**
Paste any logs you have, use CEKit verbose output: `cekit -v`.

**Environment information:**
 - Operating system: Fedora
 - CEKit version: 3.5.0

**Additional context**
Reason is that the `RedHatOverrides` include `osbs` key which is then applied as the last override. This completely removes custom `osbs` configuration.

goldmann:
Hm, you say that `RedHatOverrides` mangles the osbs section, but in fact it does not touch it at all. What `RedHatOverrides` does is to add specific environment variables and labels.

I would need to work on a reproducer first, but it's hard since I do not entirely understand the issue. Similar command specified as a reproducer is used by many teams already so it's even harder to understand the report. Can you show a working example where it does not work as expected?

goldmann:
OK, I have dig a bit more into this issue and it looks like it is in fact a bug. Most probably it was introduced by this commit: https://github.com/cekit/cekit/commit/b63e9521a4136d7a8fb70a057340e07c2d66cd2d and the changes made to the `cekit/descriptor/image.py` file. I'm looking now what is the root cause of the problem, I suspect it is related to the how the merge code works with descriptors.

goldmann:
@hmlnarik I've opened another PR that takes care of the root cause of this issue: https://github.com/cekit/cekit/pull/635

hmlnarik:
Thanks @goldmann , I meditated over removing that line with `skip_merging` as well but did have time to check all potential consequences of such a change and thus decided for a more conservative fix proposed in #632. Feel free to close that one, and thanks for fixing this bug!

PR #632, opened 2019-10-31
Do not override osbs for --redhat

hmlnarik:
Fix for #631


goldmann:
Replaced by https://github.com/cekit/cekit/pull/635.

Issue #487, opened 2019-04-17
Allow cekit  to build images from scratch


spolti:
**Describe the solution you'd like**
Discuss the possibility to allow cekit to build images from scratch;
Apparently that is not possible since during Dockerfile generation a few steps requires a user (Mostly root). i.e. when removing the /tmp/artifacts directory.

**Additional context**
Add any other context about the feature request here.

goldmann:
Thanks for this ticket, I think it should be possible to add support for building images from scratch. This could be a special case and some conditionals could definitely help.

goldmann:
I just added it to 3.1 milestone, so that it is scheduled, but I don't promise to make it part of 3.1. I think there are many corner cases that needs to be checked before we can safely say that we support it.

spolti:
Thanks for looking on this Marek, I really appreciate.
Please let me know if I can help with something else.

ppatierno:
@goldmann what's the status of this? We have to generate an image from "scratch" for the release of our product.

goldmann:
Good news, I found a solution for this RFE and I'm working now on implementing it.

ppatierno:
It's not a good news ... it's an awesome news! :-)

spolti:
Awesome Marek, thanks :)

gallettilance:
Hi all,  is there an ETA for feature completeness? where is this work being tracked?

goldmann:
Hey @gallettilance, I just came back from my PTO. This is the tracker for this work. I do not want to throw any dates here just now, but I image next 2 weeks is a possible timeframe.

PR #652, opened 2020-07-09
Add support for 'FROM scratch'

goldmann:
This commit adds support for building container images
using the special base iamge 'scratch'. Using this feature
requires that image is prepared in smart way and not every
module will work because the base image is empty and does
not provide any libraries.

A common use case for this type of images is to use it as
metadata images or to store binary builds (potentially
being a result of mult-stage builds).

To support these use cases a new property was added to resources:
'dest'. This defines where the artifact should be placed within the image.

https://github.com/cekit/cekit/issues/487

### DIFF
--- a/cekit/descriptor/resource.py
+++ b/cekit/descriptor/resource.py
@@ -5,8 +5,6 @@ import shutil
 import ssl
 import subprocess
 
-import yaml
-
 try:
     from urllib.parse import urlparse
     from urllib.request import urlopen
@@ -79,6 +77,8 @@ class Resource(Descriptor):
         self._ensure_name(descriptor)
         # Make sure the we have 'target' set
         self._ensure_target(descriptor)
+        # Add a single slash at the end of the 'dest' value
+        self._normalize_dest(descriptor)
         # Convert the dictionary into a Map object for easier access
         self._descriptor = self.__to_map(descriptor)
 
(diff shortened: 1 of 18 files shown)


goldmann:
This is initial take on it. Will probably require a few rounds of updates.

tests/images/scratch/image.yaml
```diff
@@ -0,0 +1,27 @@
...
+labels:
+  - name: "io.cekit.test"
+    value: "This is a CEKit test label"
+
+envs:
+  - name: "CEKIT_TEST"
+    value: "test"
+
+artifacts:
+    # Both files will be added to a /files/ directory in teh container (which will be created)
```
kyguy, line 15: teh -> the

goldmann: Thanks! Will fix it in the next round! :)

goldmann:
Documentation preview can be viewed at: https://1736-102005331-gh.circle-artifacts.com/0/docs/_build/html/handbook/scratch.html

cekit/descriptor/resource.py
```diff
@@ -317,6 +325,7 @@ class _PathResource(Resource):
         'map': {
             'name': {'type': 'str', 'desc': 'Key used to identify the resource'},
             'target': {'type': 'str', 'desc': 'Target file name for the resource'},
+            'dest': {'type': 'str', 'desc': 'Destination directory inside of the container, should be used together with "target"', 'default': '/tmp/artifacts/'},
```
ppatierno, line 328: the first examples in the doc don't show the usage of "target" together with "dest", or?

goldmann: Correct, `target` is optional.

goldmann: I think I could reword the description in schema. Plus I need to add this to the resource documentation, so thanks!

goldmann: This should be now resolved.

PR #645, opened 2020-03-05
Add direct support for _UrlResource in OSBS

hmlnarik:
Addresses #643

PR #661, opened 2020-11-23
CPAAS-1074 For OSBS builds use fetch-artifacts for URL based artifacts

rnc:


rnc:
Related #645

cekit/generator/osbs.py
```diff
@@ -65,23 +67,47 @@ def prepare_artifacts(self):
...
-                logger.info("Preparing artifact '{}'".format(artifact['name']))
-
-                if isinstance(artifact, _PlainResource) and \
-                        config.get('common', 'redhat'):
+                logger.info("Preparing artifact '{}' (of type {})".format(artifact['name'], type(artifact)))
+
+                if isinstance(artifact, _UrlResource):
+                    if 'md5' in artifact:
+                        md5value = artifact['md5']
+                    else:
```
goldmann, line 81: I'm not sure if I would do this. We should rely on the checksums provided by the user. In other case I could imagine two things could happen:

1. We calculate checksum for a bad artifact and pass it later to OSBS
2. Files to download could be very big.

I would just entirely skip handling URL artifacts which do not have checksums provided (required by OSBS `fetch-artifacts-url.yaml` file). I imagine these files (without checksum) should be handled similarly to what we do now (lookaside cache) since these do not fulfill all requirements to be handled differently.

Thoughts?

Side note - we should handle all checksums, not only md5, but md5, sha1, sha256 as well.

tests/test_integ_builder_osbs.py
```diff
@@ -571,7 +569,47 @@ def test_osbs_builder_with_fetch_artifacts_file_creation(tmpdir, mocker, caplog)
     assert fetch_artifacts[0] == {'md5': '123456',
                                   'target': 'artifact_name', 'url': 'http://random.url/path'}
 
-    assert "Artifact 'artifact_name' added to fetch-artifacts-url.yaml" in caplog.text
+    assert "Artifact 'artifact_name' (as plain) added to fetch-artifacts-url.yaml" in caplog.text
+
+
+def test_osbs_builder_with_fetch_artifacts_url_file_creation(tmpdir, mocker, caplog):
```
goldmann, line 575: Do we have tests that check what happens when there are other keys defined for the artifact like `target` or `dest`?

cekit/generator/osbs.py
rnc: Not sure. Looking at the descriptors, the checksums are optional with a urlartifact - so I think it makes sense to add handling for them. We only attempt to calculate if the URLArtifact hasn't got a checksum - which from cpaas they all should. 

> Side note - we should handle all checksums, not only md5, but md5, sha1, sha256 as well.

Agreed - to be updated.

tests/test_integ_builder_osbs.py
rnc: I have added more tests and for those with `target` defined. Not sure yet how to handle `dest`

cekit/log.py
```diff
@@ -26,7 +26,7 @@ def setup_logging():
     handler_err.addFilter(SingleLevelFilter(logging.INFO, True))
 
     formatter = colorlog.ColoredFormatter(
-        '%(log_color)s%(asctime)s %(name)-12s %(levelname)-8s %(message)s')
+        '%(log_color)s%(asctime)s %(filename)s:%(lineno)-10s %(levelname)-5s %(message)s')
```
rnc, line 29: While I found this *very* useful I don't know if you want this reverted. It would be nice if it could be enabled perhaps if debug mode is enabled ?

cekit/generator/osbs.py
```diff
@@ -65,26 +67,48 @@ def prepare_artifacts(self):
...
+                        logger.warning("No md5 supplied for {}, calculating from the remote artifact".format(artifact['url']))
+                        intersected_hash = ["md5"]
+                        tmpfile = tempfile.NamedTemporaryFile()
+                        try:
+                            artifact.download_file(artifact['url'], tmpfile.name)
+                            artifact["md5"] = crypto.get_sum(tmpfile.name, "md5")
+                        finally:
+                            tmpfile.close()
+
+                    fetch_artifacts_url.append({'url': artifact['url'],
```
goldmann, line 91: You don't add the computed md5 here?

goldmann: Ah, it's added later: https://github.com/cekit/cekit/pull/661/files#diff-53fb5968a14cd97636475ff271beedac8d436a8e4a42bbd44235de8e37307977R93

goldmann (approved):
LGTM, thanks!

PR #666, opened 2020-12-21
CPAAS1237 : Add description field as comment after URL

rnc:
Awaiting #661 to be merged first.


rnc:
@goldmann this also fixes the issue with broken CI on `develop` with pykwalify changing.

goldmann (approved):

PR #667, opened 2021-01-09
CPAAS1238 : Keep destination when applying override

rnc:


docs/handbook/overrides.rst
```diff
@@ -212,3 +212,68 @@ Example
...
+
+If the original contains a target definition and the override does **not** specify a target then the original value will
+be maintained.
+
+Examples
+^^^^^^^^
+
+1. Maintain destination with plain override with new target
+
+    .. code-block:: py
```
goldmann, line 233: Why not use `yaml` here as well?

docs/handbook/overrides.rst
```diff
@@ -212,3 +212,68 @@ Example
...
+
+        name: 'bar.jar'
+        dest: '/tmp/destination/'
+        md5: 234234234234
+        target: 'bar2222.jar'
+
+
+2. Maintain destination and target with plain override:
+
+    .. code-block:: py
```
goldmann, line 259: Same.

goldmann:
What happens to the `description` property? Shouldn't it be covered the same way?

docs/handbook/overrides.rst
rnc: Um. typo? Will change :-)

rnc:
> What happens to the `description` property? Shouldn't it be covered the same way?

Added coverage for that as well.

goldmann (approved):

PR #669, opened 2021-02-11
Issue #668 : Add install_weak_deps=0 to microdnf invocations

### LINKED ISSUE #668: additional arguments to package manager (specifically microdnf and --setopt=install_weak_deps=0)

I would like the invocations to microdnf generated by cekit in response to `packages:` values to include the extra command-line argument `--setopt=install_weak_deps=0`.

This causes microdnf to not install recommended or suggested RPMs ("weak" dependencies), only traditional "requires" ones. This will result in generally smaller images.

The only way to configure microdnf to do this is via command line, it is not possible to influence it via configuration file or similar: https://github.com/rpm-software-management/microdnf/blob/8a1f065966362afb96e95d177d829271a3ee3ab7/dnf/dnf-main.c#L176

**Additional context**

the last respin of the openjdk containers grew in size by 6% as a result of transitive weak package dependencies.

rnc:
Fixes #668

### DIFF
--- a/cekit/template_helper.py
+++ b/cekit/template_helper.py
@@ -90,7 +90,21 @@ class TemplateHelper(object):
 
     def default_package_manager_flags(self, pkg_mgr):
         default = "--setopt=tsflags=nodocs"
-        if "microdnf" in pkg_mgr:
+        if "apk" in pkg_mgr:
+            return ""
+        elif "microdnf" in pkg_mgr:
             return "--setopt=install_weak_deps=0 " + default
         else:
             return default
+
+    def default_package_manager_install(self, pkg_mgr):
+        if "apk" in pkg_mgr:
+            return "add"
+        else:
+            return "install -y"
+
+    def default_package_manager_query(self, pkg_mgr):
+        if "apk" in pkg_mgr:
+            return "apk info -e"
+        else:
+            return "rpm -q"
--- a/cekit/template_helper.py
+++ b/cekit/template_helper.py
@@ -88,7 +88,7 @@ class TemplateHelper(object):
         if image.get('osbs', {}).get('configuration', {}).get('container', {}).get('remote_source'):
             return True
 
-    def default_package_manager_flags(self, pkg_mgr):
+    def package_manager_flags(self, pkg_mgr):
         default = "--setopt=tsflags=nodocs"
         if "apk" in pkg_mgr:
             return ""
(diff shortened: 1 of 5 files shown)


jmtd:
I've just tested this branch with a build of [this exact image descriptor](https://github.com/jboss-container-images/openjdk/blob/aa4a4088ed51bd2df32bc381e53ffe3eaca80f98/ubi8-openjdk-8.yaml). It built fine, and the resulting image was much smaller (approx 25% ish) due to the number of RPMs that were not installed (as desired)

jmtd (approved):

jmtd (dismissed):

cekit/template_helper.py
```diff
@@ -87,3 +87,10 @@ def ports(self, available_ports):
     def cachito(self, image):
         if image.get('osbs', {}).get('configuration', {}).get('container', {}).get('remote_source'):
             return True
+
+    def default_package_manager_flags(self, pkg_mgr):
+        default = "--setopt=tsflags=nodocs"
```
goldmann, line 92: I would go one step further. Since we allow to use multiple package managers, some of them will not accept this default value (`apk` as an example). I would return empty string in case of apk, `--setopt=tsflags=nodocs` for other supported ones and add `--setopt=install_weak_deps=0`

Other thing is I don't know how did it work earlier with apk :)

goldmann (changes requested):

cekit/template_helper.py
rnc: @goldmann Perhaps apk never worked! I imagine none of the tests cover it

goldmann: Hm, we have a few tests. Needs to be checked.

rnc: @goldmann The `apk` is handled within the `template.jinja` by a separate 'if path' so that it doesn't actually use default_package_manager flags. This is why it worked before (and still does now)

goldmann: How about generalizing it into "install" and "verify" methods in helper depending on the package manager used? In other words - move code from template into helper.

goldmann: This would help with maintenance of the template and makes it handled in single place.

rnc: Updated now

goldmann (approved):

Issue #495, opened 2019-04-24
running tests requires docker (won't work with just podman or buildah etc)


jmtd:
well known issue but I couldn't find it filed already

goldmann:
Thanks for this. This will be a tricky one to solve, but at some point we need to figure out how we want to handle it. Docker, on one hand is a sensible default, but on the other hand, we should not really be tied to a particular engine.

Also, I don't think this is a bug, but it would be more an enhancement, but it depends what would be the outcome of this issue, so let's make it a discussion first and we will decide later.

rnc:
We could equally say the 'tests depend on podman being installed' - this at least would mean the user does not have to start a long-running daemon process. 

Also, slight aside - but shouldn't we clean up images after building them in tests - I see my docker image repository growing steadily?

goldmann:
Dunno really, but `docker system prune -a` is helpful to run from time to time.

Regarding Podman - I found it to be slower to build an image than Docker, haven't tried to debug it why it is like it is. Have you seen similar thing?

luck3y:
Looking at this lately, it seems close to working like a drop in for docker. If you have a recent podman and podman-docker compat installed, you can try: 

(1) chown -R root:docker run/podman
(2) chmod g+rwx /run/podman
(3) [optional if it already exists] rm /run/docker.sock
(4) ln -s /run/podman/podman.sock /run/docker.sock
Start the podman daemon:
(5) podman system service

This will likely fail when you run tests at this point becuase podman by defualt prepends localhost/ to the image names when they are built locally, which docker does not.

@goldmann Do you think changing the default podman behaviour to match docker in this case is a good idea? I don't know if it will allow an empty prefix (-t to podman i think when you build.)

spolti:
@luck3y to be honest I didn't have much time to look on this.
But so far, we didn't find a way to run everything with podman cleanly.

s2i builds is another problem, even doing s2i builds directly I didn't find a way to easly tell s2i to use podman registry instead docker.

I don't think we can do all those hacks during a productization steps.
at least for our images we still tied to docker.

rnc:
I think, going back to my comment https://github.com/cekit/cekit/issues/495#issuecomment-510451710 even if the dependency for Docker is removed, then vice versa Podman is required. For myself, while I may run _some_ tests on my local laptop I have a custom Fedora 33 VM with Docker enabled to run the complete suite on - note now enabling Docker under Fedora requires cgroup changes.  

[...]

PR #675, opened 2021-02-23
Transparently disable tests if Docker is not installed

rnc:
While running the _complete_ suite requires all engines (e.g. docker & podman) to be installed, this change is to transparently disable docker based tests if the executable is not installed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,6 @@ dist
 *.pyc
 *.swp
 target
-concreate.egg-info
 build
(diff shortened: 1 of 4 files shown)


goldmann:
@rnc I think the scope of #495 is different. Let's chat.

rnc:
I've removed the reference to 495 ; this is still relevant for running locally.

goldmann (approved):

PR #682, opened 2021-03-03
Add Python 3.8 to testing

rnc:
Adds an initial GitHub Action to run the tests.


.circleci/config.yml
```diff
@@ -49,6 +49,10 @@ jobs:
           if [[ "${CIRCLE_NODE_INDEX}" == 3 ]]; then
              make test-py37
           fi
+      - run: |
+          if [[ "${CIRCLE_NODE_INDEX}" == 3 ]]; then
```
goldmann, line 53: I think we would remove one of the Python versions from CI in case of adding 3.8.

rnc: Discussed offline - circle ci has 4 runners hence hitting the limit. Removed this change.

goldmann (approved):

PR #687, opened 2021-03-04
Process inline path overrides correctly

### LINKED ISSUE #686: Inline overrides processng
When testing #681 on devel branch I noticed  that there appeared to be problems processing inline overrides like:

```
cekit -v --descriptor image-bundle.yaml --redhat build --overrides '{"artifacts":[{"name":"kieapp.crd.yaml", "path": "deploy/olm-catalog/prod/7.11.0-1/manifests/kieapp.crd.yaml", "md5": "8c5202c91965955bafe0e143acbbd28a", "dest": "/manifests"}]}' osbs
```

rnc:
Fixes #686

### DIFF
--- a/cekit/tools.py
+++ b/cekit/tools.py
@@ -8,7 +8,7 @@ import click
 
 import yaml
 from yaml.representer import SafeRepresenter
-
+from distutils import dir_util
 from cekit.errors import CekitError
 
 try:
@@ -164,7 +164,9 @@ def copy_recursively(source_directory, destination_directory):
         if os.path.islink(src):
             os.symlink(os.readlink(src), dst)
         elif os.path.isdir(src):
-            shutil.copytree(src, dst, symlinks=True)
+            # Prefer dir_util over shutil as it doesn't throw an exception for existing
+            # directories (which we can't handle on pre Python 3.8 versions)
+            dir_util.copy_tree(src, dst, preserve_symlinks=True)
         else:
             shutil.copy2(src, dst)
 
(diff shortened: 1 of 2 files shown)


cekit/generator/base.py
```diff
@@ -53,10 +53,14 @@ def __init__(self, descriptor_path, target, overrides):
 
         if overrides:
             for override in overrides:
-                # TODO: If the overrides is provided as text, why do we try to get path to it?
                 LOGGER.debug("Loading override '{}'".format(override))
+
+                override_artifact_dir = os.path.dirname(os.path.abspath(override))
+                if not os.path.exists(override):
+                    # Just use current directory for the artifact_dir
+                    override_artifact_dir = os.getcwd()
```
goldmann, line 61: You are changing the logic here which will break runs in case the overrides in not in current directory. Saying it differently - artifact dir should be relative to the overrides file which is defining it. You are using current directory no matter where the override file is located.

Other thing is that current directory for inline overrides is probably not a good choice either. It should be directory where the image definition is located.

cekit/tools.py
```diff
@@ -38,34 +32,29 @@ def load_descriptor(descriptor):
 
     Args:
       descriptor - yaml descriptor or path to a descriptor to be loaded.
-      If a path is provided it must be an absolute path. In other case it's
-      assumed that it is a yaml descriptor.
+      In other case it's assumed that it is a yaml descriptor.
```
goldmann, line 35: This reads badly with the sentence above together.

rnc:
Closed accidentally ; submitted #688

Issue #609, opened 2019-09-03
Add support for APK package manager


cvgaviao:
**Describe the solution you'd like**
I would like to use Cekit to build images using Alpine Linux as its base image. But Alpine uses APK as its packages manager that currently is not supported by Cekit.

goldmann:
Thanks for this report! It should be doable to add support for Alpine fairly easy, I hope. Let me investigate it. I'll schedule it afterwards.

goldmann:
Scheduling for next release.

goldmann:
Sadly, I need to move this into next release. It looks like there is big demand to release 3.5 very soon and I still haven't finished the PR for this ticket. It's in works, but there are a few things that needs to be ironed still. Moving to 3.6.

goldmann:
This change will require switching from `bash` to `sh` for executing module scripts, but I don't think this will affect anyone, because in almost all cases sh is a symlink to bash.

CentOS 7:

```
[root@d82648b51043 bin]# ll /usr/bin/sh
lrwxrwxrwx. 1 root root 4 Aug  1 01:09 /usr/bin/sh -> bash
```

Fedora 30:

```
[root@fa99bdfa171a /]# ll /usr/bin/sh
lrwxrwxrwx. 1 root root 4 May 13 12:18 /usr/bin/sh -> bash
```

UBI 8:

```
bash-4.4# ls -hal /usr/bin/sh
lrwxrwxrwx. 1 root root 4 Jan 14  2019 /usr/bin/sh -> bash
```

Alpine 3.10:

```
/ # ls -hal /bin/sh 
lrwxrwxrwx    1 root     root          12 Aug 20 10:30 /bin/sh -> /bin/busybox
```

Alpine is a new kid on the block so it doesn't count :)

PR #695, opened 2021-03-17
added the ability to use Debian based distributions by adding the apt-get package manager

### LINKED ISSUE #693: Add support for APT-GET package manager

I would like to use Cekit to build images using Debian-slim as its base image.

Debian (and its derivative distributions) uses apt-get as its packages manager. 

Unfortunately apt-get is currently is not supported by Cekit.

stephengaito:
This pull request provides changes to (the development branch of) CEKit to allow the use of ``apt-get`` as a package manager.

These changes are based upon the corresponding pull request for the Alpine ``apk`` package manager #612 corresponding to the enhancement issue #609 

Following the example of the PR #612 we have added:
1. changes to the templating system (via template_helper.py),
2. updated the documentation,
3. added (minimal?) tests corresponding to the use of debian-slim and ``apt-get`` as the distribution and its package manager.

**NOTE:** The Debian ``apt-get`` package manager seems to *require* an "update" step *before* its "install" step. To be completely "correct" we *should* refactor the ``repo_install`` and ``pkg_install`` macros in the ``cekit/templates/template.jinja`` file to add an explicit "pre-install" step. However we have, for this PR, simply added ``update && apt-get --no-install-recommends`` to the ``package_manager_flags`` macro in the ``cekit/template_helper.py`` file. (see the PR differences report).

All tests pass against the Python 3.8 interpreter. (``make test``).

Fixes #693

### DIFF
--- a/cekit/descriptor/packages.py
+++ b/cekit/descriptor/packages.py
@@ -21,7 +21,7 @@ map:
   install:
     seq:
       - {type: any}
-  manager: {type: str, enum: ['yum', 'dnf', 'microdnf', 'apk']}""")
+  manager: {type: str, enum: ['yum', 'dnf', 'microdnf', 'apk', 'apt-get']}""")
 
 
 repository_schema = yaml.safe_load("""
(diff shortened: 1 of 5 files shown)


rnc (approved):

rnc:
Thanks for the contribution!

PR #698, opened 2021-03-18
Add command-line/configuration option to allow/disallow the use of ODCS content sets

### LINKED ISSUE #696: Allow the use of CEKit on host distributions which have no odcs clients

Allow the use of CEKit on host distributions (such as Debian/Ubuntu) which have no odcs clients.

Add a command-line/configuration option to allow/disallow the use of odcs content_sets (and hence the dependency check noted below).

**Additional context**

The ``Generator.dependencies`` (in the ``generator/base.py`` file) has a *hard* dependency on the ``python3-odcs-client`` python module (which, at the moment, only seems to be available on RedHat derived distributions).

On distributions which have no odcs clients *and* in CEKit descriptions which do not use ``content_sets`` this *hard* dependency stops CEKit from being used for no good reason.

stephengaito:
This pull request provides changes to (the development branch of) CEKit to add a command-line/configuration option to allow/disallow the use of ODCS content sets.

In particular *if* the ``--no-allow-odcs`` command-line switch is used (``allow_odcs=False``) then the ``Generator::dependencies`` method does *not* add the ``python3-odcs-client`` dependency and an error is raised if the CEKit descriptor makes use of ``content_sets``.

The **default** remains ``--allow-odcs`` (``allow_odcs=True``). With this default the dependency check for ``python3-odcs-client`` remains and the use of ``content_sets`` are allowed. (This default retains the current behaviour).

*These changes allow the use of CEKit on host distributions, such as Debian/Ubuntu, which do not have ODCS clients.*

Fixes #696

### DIFF
--- a/cekit/cli.py
+++ b/cekit/cli.py
@@ -26,9 +26,10 @@ default_work_dir = "~/.cekit"
 @click.option('--work-dir', metavar="PATH", help="Location of the working directory.", default=default_work_dir, show_default=True)
 @click.option('--config', metavar="PATH", help="Path to configuration file.", default=default_work_dir + "/config", show_default=True)
 @click.option('--redhat', help="Set default options for Red Hat internal infrastructure.", is_flag=True)
+@click.option('--allow-odcs/--no-allow-odcs', help="Allow ODCS content sets.", default=True)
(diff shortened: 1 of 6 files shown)


stephengaito:
After further discussion (#696) I see that this is not the correct solution.
Closing this Pull Request.

Issue #711, opened 2021-04-01
Cachito support missing WORKDIR & possibly other things


AntStephenson:
**Describe the bug**
To add support for Cachito to our Go build I added to my image.yaml:

```
 - name: "rhpam-7/rhpam-kogito-operator"
   version: "7.11"
   from: "registry.access.redhat.com/ubi8/ubi-minimal:latest"
@@ -45,6 +36,11 @@
     configuration:
       container:
         image_build_method: imagebuilder
+        remote_source:
+          repo: https://github.com/kiegroup/rhpam-kogito-operator
+          ref: db4a5d18f5f52a64083d8f1bd1776ad60a46904c
+          pkg_managers:
+            - gomod
         platforms:
           only:
             - x86_64
```

Despite this (and removing the erroneous `go mod download` ) it doesn't work. After talking with Luiz he suggested (and from the docs[1]) we are missing a WORKDIR invocation. Plus there is some other stuff in the docs too. 

**To reproduce**
See above, and execute our build with `cekit --redhat build osbs`

**Expected behavior**
I'm guessing the build should use cached artifacts from Cachito, or they should be downloaded from there during the build.

**Logs**
Paste any logs you have, use CEKit verbose output: `cekit -v`.

**Environment information:**
 - Operating system CentOS 7.9
 - CEKit version: 3.10

**Additional context**
I pasted internal links to the failed builds in the CEKit channel

[1] https://source.redhat.com/groups/public/container-build-system/container_build_system_wiki/containers_from_source_multistage_builds_in_osbs

rnc:
Need to add `WORKDIR $REMOTE_SOURCE_DIR/app`.

rnc:
Verify that remote_source is documented properly as well

rnc:
Original PR : https://github.com/cekit/cekit/pull/665
The `remote_source` is part of https://osbs.readthedocs.io/en/latest/users.html#fetching-source-code-from-external-source-using-cachito

rnc:
@AntStephenson I'm closing this as I think it was fixed in #716 and we can make further fixes in #719.

PR #716, opened 2021-04-06
Add WORKDIR for Cachito support

rnc:
For #711


rnc:
(Merging to make it easier for Alberto to test.)

PR #717, opened 2021-04-06
Add support for manager_flags

### LINKED ISSUE #714: Newer cekit versions cannot work with older microdnf due to install_weak_deps flag
I notice the latest cekit versions add the `install_weak_deps` flag [1] on the execution of a `microdnf` command.
The problem is that it cannot work for building images using an older version of microdnf which doesn't support that flag yet.
For example, I was trying to use cekit 3.9 to build an image which uses as base ubi7minimal and when running, cekit raises this error:

```shell
2021-04-03 15:40:13,423 docker_builder.py:108        INFO  Docker: Step 7/21 : RUN microdnf --setopt=install_weak_deps=0 --setopt=tsflags=nodocs install -y unzip tar rsync shadow-utils             && microdnf clean all             && rpm -q unzip tar rsync shadow-utils
2021-04-03 15:40:13,424 docker_builder.py:108        INFO  Docker: 
2021-04-03 15:40:13,562 docker_builder.py:108        INFO  Docker: ---> Running in e7aecd6d7100
2021-04-03 15:40:13,975 docker_builder.py:108        INFO  Docker: error: (--setopt) Unable to handle: install_weak_deps=0
```
Running the same command line on a ubi8minimal image works fine.
So I noticed that ubi7minimal has `microdnf-2-8.el7.x86_64` while ubi8minimal has `microdnf-3.4.0-1.el8.x86_64`; only the newer supports the `install_weak_deps` flag.
A colleague run the build using an older cekit (3.7) and it worked because it generatesd the above microdnf command in the Docker image file but without the above flag.

I was wondering if there is the possibility to disable flags for microdnf through the image descriptor; I couldn't find anything useful in the documentation but just the note "For microdnf, the flag --setopt=install_weak_deps=0 is also added.".

[1] https://github.com/cekit/cekit/blob/develop/cekit/template_helper.py#L109

rnc:
Fixes #714

### DIFF
--- a/cekit/template_helper.py
+++ b/cekit/template_helper.py
@@ -94,11 +94,13 @@ class TemplateHelper(object):
     def extra_dir_target(self, image):
         return image.get('osbs', {}).extra_dir_target
 
-    def package_manager_flags(self, pkg_mgr):
-        default = "--setopt=tsflags=nodocs"
-        if "apk" in pkg_mgr:
+    def package_manager_flags(self, pkg_mgr, pkg_mgr_flags):
+        if pkg_mgr_flags is not None:
+            # Using None check to allow a definition of "manager_flags: ''" to override the default values.
+            return pkg_mgr_flags
+        elif "apk" in pkg_mgr:
             return ""
-        if "apt-get" in pkg_mgr:
+        elif "apt-get" in pkg_mgr:
             #
             # This is a HACK...
             #
@@ -111,7 +113,9 @@ class TemplateHelper(object):
             # However this works at the moment...
             #
             return "update && apt-get --no-install-recommends"
-        elif "microdnf" in pkg_mgr:
+
+        default = "--setopt=tsflags=nodocs"
+        if "microdnf" in pkg_mgr:
             return "--setopt=install_weak_deps=0 " + default
         else:
             return default
--- a/cekit/descriptor/packages.py
+++ b/cekit/descriptor/packages.py
@@ -21,7 +21,8 @@ map:
   install:
     seq:
       - {type: any}
-  manager: {type: str, enum: ['yum', 'dnf', 'microdnf', 'apk', 'apt-get']}""")
+  manager: {type: str, enum: ['yum', 'dnf', 'microdnf', 'apk', 'apt-get']}
+  manager_flags: {type: str}""")
 
 
 repository_schema = yaml.safe_load("""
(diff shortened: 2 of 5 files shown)


cekit/template_helper.py
```diff
@@ -111,7 +113,9 @@ def package_manager_flags(self, pkg_mgr):
             # However this works at the moment...
             #
             return "update && apt-get --no-install-recommends"
-        elif "microdnf" in pkg_mgr:
+
+        default = "--setopt=tsflags=nodocs"
+        if "microdnf" in pkg_mgr:
```
ppatierno, line 118: not a Python developer but as far as I understood if no manager flags are specified and the manager is microdnf, the `--setopt=install_weak_deps=0` is set by default ... but it's exactly what brakes backward compatibility related to #714, or? What am I missing?

rnc: This is still to be debated with @jmtd and @goldmann but this PR allows user's who are on latest release and using ubi8 to have the preferred options by default, with those needing to have backwards compatibility can override as desired.

ppatierno: tbh, I think that the idea of backward compatibility is doing nothing for people using older stuff not the need to override something to make it working again. The newer cekit version is breaking backward compatibility actually.

goldmann (approved):

ppatierno:
@goldmann you merged this, so it means that backward compatibility with ubi7 is finally broken because cekit is going to add the `--setopt=install_weak_deps=0` by default which doesn't work?

goldmann:
@ppatierno The "breaking" change was part of 3.9, not this particular PR.

We had to choose one option out of the two available: support latest version or support most generic/common and I'm not sure if one option is better than the other. The only thing I know is that this will make some people unhappy for sure. But we had to decide.

In general, after upgrading CEKit version (which is not a bugfix version) you should expect some changes. And this is was the case for 3.9 release regarding the `microdnf` package manager flag. Now (in this PR) we are not changing it again and instead - making it possible to override it, keeping the default as the most optimized for the latest version.

This means that this PR adds an option to adjust the flags, if defaults are not the best ones which is the case for UBI 7 images, as you mentioned. Docs: https://2129-102005331-gh.circle-artifacts.com/0/docs/_build/html/descriptor/image.html#package-manager-flags

Hope this helps!

(I opened #724 about failures on docs)

PR #739, opened 2021-07-21
Docs: Include all keys in top-most OSBS example

jmtd:
Include an example of every (sub) key for OSBS configuration in
the example YAML at the top of the page.

It was not clear enough to me where in the osbs heirarchy the
`koji_target` key belonged from the documentation alone, and I
had to double check by reading the source code.

I've ordered the keys to match the order of descriptions below.

Signed-off-by: Jonathan Dowland <jdowland@redhat.com>

### DIFF
--- a/docs/descriptor/includes/image/osbs.rst
+++ b/docs/descriptor/includes/image/osbs.rst
@@ -9,9 +9,12 @@ Required
 This section represents object we use to hint OSBS builder with a configuration which needs to be tweaked
 for successful and reproducible builds.
 
-It contains two main keys:
+It contains five keys:
 
+* :ref:`extra_dir <descriptor/image:OSBS extra directory>`
+* :ref:`extra_dir_target <descriptor/image:OSBS extra copy command>`
 * :ref:`repository <descriptor/image:OSBS repository>`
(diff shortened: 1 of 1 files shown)


rnc:
@jmtd Can you rebase please (I've pushed #740 to solve the test failure)

docs/descriptor/includes/image/osbs.rst
```diff
@@ -18,10 +18,14 @@ It contains two main keys:
 .. code-block:: yaml
```
rnc, line 18: The section above, `It contains two main keys:....` I think is also incorrect as OSBS contains 5 keys not 2. Would it be possible to amend the above as well please?

jmtd: Sure, done. I've added inter-links for all five sections too, underneath.

jmtd: I'd be grateful if someone could verify the added links are OK (I tried rendering the page using pandoc but it did not process the links)

rnc: Links look fine - the `make preview` in the docs subdirectory helps to render them :-)

rnc (approved):

PR #744, opened 2021-08-12
Allow different checksum types for PlainArtifacts. For YAML loading w…

### LINKED ISSUE #741: artifact handling: md5 sums special-cased, other sum types not supported for cache copy
Say you have an artifact definition as follows

```
artifacts:
- name:   java-17-openjdk-headless-17
  target: java-17-openjdk-headless-17.0.0.0.26-0.1.ea.el8.x86_64.rpm
  sha256: 2751509558cef22e3c69aae4414521946a7fff2c0b35c4e225895dea2bf52ea7
```

Cekit (3.11.0) will fail the build with

    ERROR Resource '{'name': 'java-17-openjdk-headless-17', 'target': 'java-17-openjdk-headless-17.0.0.0.26-0.1.ea.el8.x86_64.rpm', 'sha256': '2751509558cef22e3c69aae4414521946a7fff2c0b35c4e225895dea2bf52ea7'}' is not supported

The issue is that when none of 'image','path','url' or 'git' keys are specified, `create_resource` has [one more logic check for the `md5` key](https://github.com/cekit/cekit/blob/ca7384c7f8f902c2994a684dd0241a2fde52094d/cekit/descriptor/resource.py#L55) before failing with the above message.

1. that branch should be updated to check for any of the supported hash types
2. the error message should be rewritten to be clearer as to what the problem is

The work-around in my case is to add `md5` keys.

rnc:
…arning

Fixes #741 

@jmtd  Can you verify this is what you wanted please?


goldmann:
Hm, we base on md5 for plain artifacts because the `brew listArchives` command supports md5 only from what I know. If we change the requirement - we may not be able to fetch any artifacts if we change this. This could lead to unexpected behaviors where we would need to debug failures like: "I did specify correct checksum, but build failed - why?".

rnc:
> Hm, we base on md5 for plain artifacts because the `brew listArchives` command supports md5 only from what I know. If we change the requirement - we may not be able to fetch any artifacts if we change this. This could lead to unexpected behaviors where we would need to debug failures like: "I did specify correct checksum, but build failed - why?".

Yeah I think you're right and we need an extra check - for this, it should be possible to use a PlainArtifact with any checksum and _only_ if md5 **and** redhat flags are available could the brew lookup be done.

goldmann (approved):

Issue #78, opened 2017-09-29
Suggestion: show some kind of download progress feedback 


jrenaat:
When doing generate for large source files, the console doesn't show any feedback on progress, making it hard to establish whether the tool is working, or suffering some kind of problem instead.

rnc:
@jrenaat Do you mean when CEKit downloads large remote URL based resources or some other action?

jrenaat:
After 4 years, I really don't remember what I meant exactly ;)

rnc:
Yeah sorry about that ; I've started going over older issues before I got involved :-)

rnc:
Fixed with #751

PR #751, opened 2021-09-07
Add progress bar for remote downloads

### LINKED ISSUE #482: Add progress report when adding files to cache using cekit-cache command
It would be great to have some progress bar or something like this.

**Additional context**
None.

rnc:
Fixes #482, #78

### DIFF
--- a/cekit/descriptor/resource.py
+++ b/cekit/descriptor/resource.py
@@ -4,6 +4,9 @@ import os
 import shutil
 import ssl
 import subprocess
+import sys
+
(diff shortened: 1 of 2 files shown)


goldmann:
How does this handle scenario when the lengths is unknown? What does it print?

rnc:
If the Content-Length field is missing , it defaults to zero which means the bar is then 100% after first chunk (so they still get feedback).

goldmann (approved):

PR #753, opened 2021-09-08
Support remote URLs for overrides-file command

### LINKED ISSUE #627: Add support for fetching override files from url
It should be possible to run:

```
cekit build --overrides-file http://host.com/path/to/overrides.yaml podman
```

rnc:
Fixes #627

### DIFF
--- a/cekit/generator/base.py
+++ b/cekit/generator/base.py
@@ -5,21 +5,26 @@ import os
 import platform
 import re
 import shutil
+import tempfile
 
 from jinja2 import Environment, FileSystemLoader
 from packaging.version import LegacyVersion, parse as parse_version
 
-from cekit import tools
 from cekit.config import Config
 from cekit.descriptor import Env, Image, Label, Module, Overrides, Repository
 from cekit.errors import CekitError
 from cekit.template_helper import TemplateHelper
 from cekit.version import __version__ as cekit_version
+from cekit.tools import download_file, load_descriptor
 
 LOGGER = logging.getLogger('cekit')
 CONFIG = Config()
 
 
+try:
+    from urllib.parse import urlparse
+except ImportError:
+    from urlparse import urlparse
 try:
     from odcs.client.odcs import ODCS, AuthMech
     # Requests is a dependency of ODCS client, so this should be safe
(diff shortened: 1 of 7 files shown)


cekit/generator/base.py
```diff
@@ -60,11 +61,18 @@ def __init__(self, descriptor_path, target, overrides):
 
                 LOGGER.debug("Loading override '{}'".format(override))
 
-                override_artifact_dir = os.path.dirname(os.path.abspath(override))
-                if not os.path.exists(override):
-                    override_artifact_dir = os.path.dirname(os.path.abspath(descriptor_path))
-                self._overrides.append(Overrides(tools.load_descriptor(
-                    override), override_artifact_dir))
+                # HTTP Handling
+                if override.startswith('http'):
```
goldmann, line 65: I wonder if we need to do this if we have the `urlparse()` mechanism which provides `scheme`.

rnc: Yeah I think its a toss-up which to go with:
```
parsed_url = urlparse(url)
if parsed_url.scheme in ['http', 'https']:
```
or
```
if override.startswith('http'):
```

Actually one potentially benefit could be to handle `file://` URLs as well so if we do:
```
parsed_url = urlparse(url)
if parsed_url.scheme:
```

goldmann: Yah, I think it would make sense to refactor it to use the same mechanism. Or just use the `download` (possibly rename?) method every time. But this would require some more refactoring I guess.

rnc: Done

cekit/generator/base.py
```diff
@@ -60,11 +62,18 @@ def __init__(self, descriptor_path, target, overrides):
 
                 LOGGER.debug("Loading override '{}'".format(override))
 
-                override_artifact_dir = os.path.dirname(os.path.abspath(override))
-                if not os.path.exists(override):
-                    override_artifact_dir = os.path.dirname(os.path.abspath(descriptor_path))
-                self._overrides.append(Overrides(tools.load_descriptor(
-                    override), override_artifact_dir))
+                if urlparse(override).scheme:
```
goldmann, line 65: Isn't here missing something? Like the scheme comparison?

rnc: I was relying on the fact, if the scheme is filled in by `urlparse` and not an empty string, then download_file can handle the rest ; if its an unsupported scheme then download_file will detect that and throw an exception.

**However**, saying that it looks like `urlparse` gets really confused when it sees e.g. `labels:` and assumes that is a URL scheme (it mustn't check for `//`) .

goldmann (approved):

PR #772, opened 2022-03-29
Refactor and improve logging for CalledProcessError

rnc:


cekit/cli.py
```diff
@@ -462,6 +463,13 @@ def run(self, clazz):
             sys.exit(0)
         except KeyboardInterrupt:
             pass
+        except CalledProcessError as ex:
+            LOGGER.error(
```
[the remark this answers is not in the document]
rnc: The first `{}` is specially handled by `CalledProcessError`. An example:
```
ERROR Command 'git' returned non-zero exit status 1. Command stdout is 'A GIT ERROR' with stderr 'None'
```

cekit/builders/osbs.py
```diff
@@ -354,7 +358,12 @@ def run(self):
                     task_id = subprocess.check_output(cmd).strip().decode("utf8")
 
                 except subprocess.CalledProcessError as ex:
-                    raise CekitError("Building container image in OSBS failed", ex)
+                    LOGGER.error(
+                        "{} Command stdout is '{}' with stderr '{}'".format(
+                            ex, ex.stdout, ex.stderr
+                        )
+                    )
+                    raise CekitError("Building container image in OSBS failed")
```
rnc, line 366: Pros/cons here ; while we do get a nice CekitError bubbling up ; it would be equally possible to remove the try/catch and just let the CalledProcessError bubble all the way up

cekit/cli.py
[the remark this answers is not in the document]
goldmann: Right, what I'm looking after is basically having a separate, probably 3 calls in this case, to the `LOGGER.error`.

What you posted above is fine, for simple cases. I imagine that there will be cases where a pretty long, multi-line error will be passed and the formatting of it this way could lead to problems understanding what is the original one vs. what was added by us vs. what is still part of stdout and what is stderr now.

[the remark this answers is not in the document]
goldmann: Potentially yes. At first I was thinking that why to log empty stuff, but I guess in case of failures, making it explicit that the output was empty is a good thing.

I imagine that it could be like: `"Standard output:\n{}"`. Then we know that the content below the line is original. Similar approach to error.

cekit/builders/osbs.py
rnc: I think I'll remove this particular try/catch.

docs/requirements.txt
```diff
@@ -1,4 +1,4 @@
-sphinx>=1.7.6,<3.0
+sphinx>=1.7.6
```
rnc, line 1: Alternative to this change (and the `schema.py` above is to pin e.g.
```
Jinja2==2.11.3
markupsafe==2.0.1
```

rnc: With either the pinned version or this current one I do see
```
/home/rnc/Work/CEKit/GitHub/cekit/docs/descriptor/includes/image/packages.rst:1: WARNING: Problems with "include" directive path:
InputError: [Errno 2] No such file or directory: 'descriptor/includes/image/includes/packages/common.rst'.
```

but the doc generation (and on the PR) looks ok.

rnc: Ok so in that case going with what we have in this PR is better :-)

PR #774, opened 2022-04-06
Resolve issues with Docker API version

### LINKED ISSUE #773: Invalid docker version reported by cekit
**Describe the bug**
I am using cekit 4.1.0 and docker builder. The docker client and server version are 1.41. I do cekit build docker --platform linux/amd64, cekit complains that: ERROR ('Image build failed, see logs above.', InvalidVersion('platform was only introduced in API version 1.32')). 

**To reproduce**
cekit build docker --platform linux/amd64

**Expected behavior**
Multi arch should be built.

**Logs**
Paste any logs you have, use CEKit verbose output: `cekit -v`.

**Environment information:**
 - Operating system: ubuntu
 - CEKit version: 4.1.0

**Additional context**

rnc:
Fixes #773

### DIFF
--- a/cekit/builders/docker_builder.py
+++ b/cekit/builders/docker_builder.py
@@ -195,7 +195,7 @@ class DockerBuilder(Builder):
                 )
             )
 
-        params = {"version": "1.22"}
+        params = {"version": "1.35"}
         params.update(docker.utils.kwargs_from_env())
         params["timeout"] = timeout
 
--- a/tests/images/multi-stage/modules/build/module.yaml
+++ b/tests/images/multi-stage/modules/build/module.yaml
@@ -3,8 +3,8 @@ version: 1.0
 
 artifacts:
   - name: ant.tar.gz
-    url: https://downloads.apache.org//ant/binaries/apache-ant-1.10.11-bin.zip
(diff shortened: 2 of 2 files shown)


cekit/builders/docker_builder.py
```diff
@@ -195,7 +195,7 @@ def _docker_client(self):
                 )
             )
 
-        params = {"version": "1.22"}
+        params = {}
```
goldmann, line 198: Wouldn't this cause problems when the Docker API of newer version would break our usage of it?

rnc: Even older versions of the library (e.g. 4.0.2 ; https://github.com/docker/docker-py/blob/4.0.2-release/docker/api/client.py#L84 )  ( we're currently on 5.0.3 ) have 1.35 as the default. The problem occurred because we were setting it to a *really* old version.

rnc: In fact, its been 1.35 for 4 years ( https://github.com/docker/docker-py/commit/abd60aedc7e3df813006919222d86717eb8c6fc2 )

rnc: Although that appeared to cause test failures ; experimenting with updating the API version instead

rnc: @goldmann I do wonder if we should document the API version we support?

goldmann (approved):

cekit/builders/docker_builder.py
goldmann: We probably could, but I'm not sure if this is out of interest users since we pretty much hide it when using the Docker builder.

PR #783, opened 2022-05-25
Encapsulate multi-stage builds into modules

greenatatlassian:
This PR shows an example of how we might improve encapsulation of modules.

The documentation for [multi-stage builds](https://docs.cekit.io/en/latest/handbook/multi-stage.html) shows how CEKit supports multi stage builds by allowing a module to contain an artifact that references a previous build image. To enable this functionality, it appears that the image descriptor was modified to enable a list of images descriptors in the file so that "build" stages could be defined.

Having multi-stage builds is really nice, but I feel the current design here breaks the encapsulation that modules might otherwise offer. A better design would allow me to list a module I want installed in my image, and not need to worry about the details of how that module is implemented—i.e. not need to provide the necessary builder image in my image descriptor to make the module function correctly. 

In this PR, I extend the existing multi-stage functionality to allow modules themselves to define the builder image required to build some artifact, so that the image descriptor can merely reference the module to be installed. 

The test case that I've added is somewhat contrived, but hopefully shows the idea. Because CEKit's image descriptor doesn't offer any way to add `RUN` commands to a docker file directly (and I don't disagree with this choice), it is necessary to nest another module inside the first in order to run the `unzip.sh` script in the builder image. 

This PR was a quick attempt to see how hard this would be to get working, and isn't really polished (e.g. doesn't include updated documentation). If the maintainers feel like this is a step in the right direction, then I *may* be able to polish this up.


greenatatlassian:
Sorry about the failing tests. I'll check what all is being run and get everything passing sometime next week.

Issue #780, opened 2022-05-23
No tests from repo root or modules are run


greenatatlassian:
**Describe the bug**
Test files from `tests` directory next to the module descriptor are ignored.

The problem is that the code in `collector.py` copies from the target directory, `target/images/modules/<module_name>/tests/*` as confirmed by `test_collect_test_from_module`. However, this directory has already been cleaned up by `cli.cleanup` which is called when the application starts up.

**Expected behavior**
I would have expected the tests to be copied into the target directory and run along side tests from other locations as part of `cekit test behave`.

**Logs**
```
> cat image.yaml 
───────┬────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: image.yaml
───────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ schema_version: 1
   2   │ 
   3   │ name: "dummy"
   4   │ version: "1"
   5   │ description: "Golden Image with OS base layer for development based on Debian."
   6   │ 
   7   │ from: "debian:bullseye-slim"
   8 + │ 
   9 + │ modules:
  10 + │   repositories:
  11 + │     - path: "./modules"
  12 + │ 
  13 + │   install:
  14 + │     - name: "mymodule"
───────┴────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

 > tree .
.
├── image.yaml
└── modules
    └── mymodule
        ├── module.yaml
        └── tests
            ├── features
            │   └── mymodule.feature
            └── steps
                └── steps.py

5 directories, 4 files

 > cat modules/mymodule/tests/features/mymodule.feature 
───────┬────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: modules/mymodule/tests/features/mymodule.feature
───────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ Feature: my module
   2   │ 
   3   │   @dummy
   4   │   Scenario: Test features in module
   5   │     Given behave test framework with custom test
   6   │     Then test passes
───────┴────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

 > cekit build docker
[...]

PR #784, opened 2022-06-02
Fix for #780.

greenatatlassian:
This PR provides a fix for #780.

Basically, the setup for tests was incomplete, with part of the necessary setup that is done for builds not being done for tests. The actual fix is moving one method call in `generate/base.py` (contained in the commit "Make target dir prep consistent").

I've also included a few code cleanups, and a test showing that this fixes the issue.


rnc:
Thanks for the PR (and especially for providing a unit test)! I agree it looks like its the `copy_modules` which is part of `generate`. causes the issue. I am wondering whether it would be more appropriate to add a simple Generator implementation e.g. 
```
class BehaveGenerator(Generator):
    def __init__(self, descriptor_path, target, overrides):
        super(BehaveGenerator, self).__init__(descriptor_path, target, overrides)

    def prepare_artifacts(self):
        pass
```
as that would then avoid moving the `copy_modules` into the init code. I am not sure why the code is like this as it dates from when @goldmann and @jmtd worked upon it.

Having experimented locally there are some small implications - namely
* `_prepare_repository` could be defaulted in the base class.
*  `test_simple_image_test` needs some small modifications as then the Dockerfile would be generated.

Further, from looking at the CI, I think some MacOS exclusions would be helpful.

If you have no objections, I'll make those small changes and push an update to this PR.

rnc (dismissed):

rnc (approved):

rnc (approved):

greenatatlassian:
Hi @rnc, yes, I think this is a good solution! Thank you!

PR #787, opened 2022-06-21
Add support for follow_tag

### LINKED ISSUE #760: Support automatic update when using a floating image tag to a static tag
Submitted originally as ENTMQBR-5586 by @brusdev

To keep container builds updated to the latest parent (base) images projects typically use floating (rolling) image tags. But there is an issue with this approach - it leads to non-reproducible builds and makes it impossible to validate all the builds in a release that are based on the same parent image. Red Hat products require that builds be reproducible. To address this issue, CPaaS can use static image tags which are automatically updated when the floating tags change. See CPaaS Midstream functionality with "container parent image updates"

rnc:
Signed-off-by: Nick Cross <ncross@redhat.com>

Fixes #760


brusdev:
I tried it locally adding the `follow_tag` in the image file or in the overrides file. It works like a charm!!!

rnc:
I've moved the Python 3.6 and UTF-8 normalisation to separate PRs (#788 and #789) and will commit the subprocess changes in a separate branch (https://github.com/rnc/cekit/tree/SUBPROCESS)

PR #788, opened 2022-06-24
Deprecate support for Python 3.6

rnc:


rnc:
This PR is being reverted in #793

PR #791, opened 2022-06-28
Install modules (including executions) after packages are installed

### LINKED ISSUE #790: Docker file ordering prevents modules from using installed packages in their execute scripts.
**Describe the bug**
I have a module that will be installed in an image. That module includes an `execute` script that makes use of some packages that should be installed in the image (e.g. fortune). In my image descriptor, I include those packages to be installed, and the module. However, docker build fails because the package required by the execute script is not installed in the image.

**To reproduce**
Image descriptor:
```yaml
from: alpine:latest
modules:
  install:
  - name: test_module
  repositories:
  - name: modules
    path: modules/
name: test_image
packages:
  install:
  - fortune
  manager: apk
version: '1.0'
```
Module descriptor:
```yaml
name: test_module
version: 1.0
install:
  - script: run_fortune.sh
```
`run_fortune.sh` install script:
```bash
#!/bin/bash
fortune
```

**Expected behavior**
I would expect that packages listed in the `packages` section of the image descriptor would be installed *before* any *execute* steps from modules are run.

greenatatlassian:
This PR should resolve #790.

Basically, the Jinja2 template puts the module instructions in the docker file before the package install instructions, so if the module includes an execute script (which will be run by `RUN ...`), this script will not have access to any of the packages listed in the image descriptor to be installed.

This PR adds a test (which fails without the accompanying code change), and moves the relevant line of the Jinja2 template up to get the packages installed before the modules are handled.

Other changes are to change the ordering in expected docker files to match actual docker files for some other tests.

### DIFF
--- a/tests/test_integ_install_ordering.py
+++ b/tests/test_integ_install_ordering.py
@@ -1,5 +1,7 @@
+import platform
 import sys
 
+import pytest
 import yaml
 from click.testing import CliRunner
 
@@ -23,6 +25,9 @@ def run_cekit(image_dir, descriptor, args=None):
         return result
 
 
+@pytest.mark.skipif(
+    platform.system() == "Darwin", reason="Disabled on macOS, cannot run Docker"
+)
 def test_module_uses_installed_package_in_execute_script(tmp_path):
     """Check that when a module has an install script that depends on packages that should be installed in an image,
     the install script succeeds.


greenatatlassian:
Hrm. The tests are failing because they cannot connect to the docker daemon in macOS. Perhaps this is a temporary failure in CircleCI?

rnc:
I think its because Docker is normally disabled on mac https://github.com/cekit/cekit/blob/develop/tests/test_validate.py#L345 (a lot of the other tests use mocker to avoid that)

rnc:
Thanks, this looks fine to me, I think we just need to skip on Mac.

greenatatlassian:
Sounds good. I'll plan to fix this up early next week.

tests/test_integ_install_ordering.py
```diff
@@ -0,0 +1,64 @@
+import sys
```
rnc, line 1:
```suggestion
import platform
import sys
```

rnc (approved):

PR #806, opened 2022-07-27
Revert changes for module processing order.

### LINKED ISSUE #805: Image Descriptor and Module Ordering
Within #790 an issue was raised about Dockerfile ordering on generation. Having had a bug reported that recent builds have been broken I have taken another look at this.

The issue is that the image descriptor generation ordering is **not** well documented and only the module generation is documented in https://docs.cekit.io/en/latest/handbook/modules/merging.html#module-processing-in-template In actual fact, the descriptor generation is very similar:

This was the order **prior** to the change from #790:

- from generation
- helper directory handling
- process image
   - repository processing
   - loop through all modules for processing
      - See https://docs.cekit.io/en/latest/handbook/modules/merging.html#module-processing-in-template for details here
   - for this image, perform:
      - package installation
      - env vars
      - labels
      - ports
      - executions
      - volumes

The change in 790 moved module processing to the end. I have checked back to prior to when multi-stage builds were added (which was a significant change in 2019) and that would be https://github.com/cekit/cekit/blob/1ae1708ecdce6eab69fb21b4fa51243b0fd43e54/cekit/template_helper.py - the same ordering applies here i.e. all modules are processed before the containing image is processed.

On reflection I am beginning to think a better solution for #790 would have been to have moved the package installation into the module.

rnc:
Fix handling of package managers in included repositories. Improve documentation.

Revert "Install modules (including executions) *after* packages are installed, not before."
This reverts commit e9410afcadd13a13a7d8a57e23ce6666ec038ecc.

Fixes #805

### DIFF
--- a/cekit/descriptor/image.py
+++ b/cekit/descriptor/image.py
@@ -388,6 +388,7 @@ class Image(Descriptor):
         self.packages._descriptor["repositories"] = list(
             self._package_repositories.values()
         )
+
         # final 'run' value
         if self.run:
             self.run = self.run.merge(self._module_run)
@@ -396,7 +397,7 @@ class Image(Descriptor):
 
     def process_install_list(
         self, source, to_install_list, install_list, module_registry
-    ):
+    ) -> None:
         module_overrides = self._image_overrides["modules"]
         artifact_overrides = self._image_overrides["artifacts"]
         for to_install in to_install_list:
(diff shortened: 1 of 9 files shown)


jmtd (approved):
LGTM. My images seem to generate correctly. I haven't tested the troublesome Mandrel ones, should we ask them to try this?

zakkak:
> I haven't tested the troublesome Mandrel ones, should we ask them to try this?

Thank you @rnc. I confirm that the issue with the Mandrel images is fixed in CEKIT 4.3.0

PR #814, opened 2022-10-27
Issue-813: Stripping package version in package listing command

### LINKED ISSUE #813: Installation of any package with a specific version through `apt-get` in the CEKit modules fails
**Describe the bug**
Installation of any package with a specific version through `apt-get` in the CEKit modules fails.

**To reproduce**

1. Create a module (`module.yaml`)
2. Edit that module and add `packages` -> `install` section
3. Use the package manager `apt-get`
4. List any package for installation but with a specific version, for ex. `java-17-amazon-corretto-jdk=1:17.0.4.9-1`
5. My complete `module.yaml` looks like,
```
schema_version: 1

name: install.jdk
version: 1.0
description: "Installs the mentioned version of Java-17 JDK"

packages:
  manager: apt-get
  install:
    - java-17-amazon-corretto-jdk=1:17.0.4.9-1
```

**Expected behavior**
CEKit should be able to build a docker image in which `java-17-amazon-corretto-jdk`  with version `1:17.0.4.9-1` is installed.

**Logs**
```
cekit --version
4.3.0
```

```
cekit build docker

[error] dpkg-query --list java-17-amazon-corretto-jdk=1:17.0.4.9-1' returned a non-zero code: 1
```

**Environment information:**
 - Operating system : macos, linux
 - CEKit version: 4.3.0

**Additional context**
The issue occurs in the step where CEKit adds an additional command for listing the installed packages (i.e. `dpkg-query`) after each installation. See [line #139](https://github.com/cekit/cekit/blob/develop/cekit/template_helper.py#L139). If the package is listed with a specific version (for ex. `xyz=1.0.0`) in the `module.yaml` then `dpkg-query` is trying to find the package along with the version, which it doesn't find and throws an error. The simple fix is to strip the version before giving the input to `dpkg-query`. The fix will go here at [line #46](https://github.com/cekit/cekit/blob/develop/cekit/templates/template.jinja#L46).

ankithbtik:
Fixes #813

If the installable package contains the version, then the package listing command (`dpkg-query --list`) throws an error because it appends the version with the package name in the search query.
The error looks like this,
`dpkg-query --list xyz=1.x.y' returned a non-zero code: 1`

This PR fixes that issue simply by stripping the version part and using only the package name as the input for `dpkg-query --list` command.

### DIFF
--- a/cekit/templates/template.jinja
+++ b/cekit/templates/template.jinja
@@ -43,7 +43,7 @@ rm{% for repo in repositories %} /etc/yum.repos.d/{{ repo.filename }}{% endfor %
     {% if pkg_manager == 'microdnf' %}
             && {{ pkg_manager }} clean all \
     {% endif %}
-            && {{ helper.package_manager_query(pkg_manager) }}{% for package in packages %} {{ package }}{% endfor %}
+            && {{ helper.package_manager_query(pkg_manager) }}{% for package in packages %} {{ package.split('=')[0] }}{% endfor %}
     {% endif %}
 {%- endmacro -%}
 
--- a/tests/test_dockerfile.py
+++ b/tests/test_dockerfile.py
@@ -541,15 +541,15 @@ def test_supported_package_managers_apt(tmpdir, caplog):
         descriptor={
             "packages": {
                 "manager": "apt-get",
-                "install": ["a"],
+                "install": ["a", "b=1.0.0"],
                 "repositories": [{"name": "foo", "rpm": "foo-repo.rpm"}],
             }
         },
     )
     regex_dockerfile(
-        target, "RUN apt-get update && apt-get --no-install-recommends install -y a"
+        target, "RUN apt-get update && apt-get --no-install-recommends install -y a b=1.0.0"
     )
-    regex_dockerfile(target, "dpkg-query --list a")
+    regex_dockerfile(target, "dpkg-query --list a b")
     assert (
         "Package manager apt-get does not support defining repositories, skipping all repositories"
         in caplog.text
--- a/tests/test_dockerfile.py
+++ b/tests/test_dockerfile.py
@@ -547,7 +547,8 @@ def test_supported_package_managers_apt(tmpdir, caplog):
         },
     )
     regex_dockerfile(
-        target, "RUN apt-get update && apt-get --no-install-recommends install -y a b=1.0.0"
+        target,
+        "RUN apt-get update && apt-get --no-install-recommends install -y a b=1.0.0",
     )
     regex_dockerfile(target, "dpkg-query --list a b")
     assert (


jmtd:
I don't particularly like this change, I'd prefer an extention of cekit to recognise package version as outlined in my comment here: https://github.com/cekit/cekit/issues/813#issuecomment-1293196652

For Debian-based systems, this change is OK, because '=' is guaranteed not to be legal in a Debian package name ([c.f.](https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-source)).

I've no idea about the other packager managers (e.g. can an RPM package name contain a '=' character?) and this code is not specific to a given package manager.

ankithbtik:
Yes, I made an assumption that other package managers also don't allow putting `=` in the package name. 
However, it is a trivial change to make this code specific to `apt-get`. Happy to make that change if you agree.

goldmann:
I don't thin the equal sign is allowed in a package name in any of the package managers which makes the proposed solution not breaking anything. Am I wrong? I think @rnc is doing some research now.

rnc:
I think to install specific versions of a package, for:

* apk its `<package>=<version>`
* apt-get (dpkg-query) its `<package>=<version>`
* yum its NEVRA, NEVR, NVR e.g. `<package>-<version>`
* microdnf/dnf it can be NEVRA, NEVR, NVR e.g. `<package>-<version>-<release>`

Fedora naming guidelines (like Debian in https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-source ) also don't have `=` as allowed in a package name (https://fedoraproject.org/wiki/Packaging:Naming). Closest I've found for Alpine is https://wiki.alpinelinux.org/wiki/Package_policies 

For the latter two, its possible to query on package name, name+version or name+version+release. For the former two, I _think_ its only possible to query on package name.

Therefore I think this should actually work for all cases. It would however be better to add (or adapt existing e.g. `test_dockerfile.py`) an integration test :-)

ankithbtik:
Sure. I extended an existing test (written for `apt-get`) to install packages with and without the versions and then check the Dockerfile commands for installing and listing those packages.

rnc (approved):

ankithbtik:
As this is the first time I am making changes in CEKit, so I am a little unsure about the merge and release process. Will someone from the authorized persons (as I am not authorized to merge the PR) take care of merging this PR and releasing the changes? Thanks.

goldmann:
@ankithbtik Correct, someone from our team will take care of everything!

P.S. Thanks for contribution!

goldmann (approved):

rnc:
> As this is the first time I am making changes in CEKit, so I am a little unsure about the merge and release process. Will someone from the authorized persons (as I am not authorized to merge the PR) take care of merging this PR and releasing the changes? Thanks.

I'll merge it today :-) Thanks for the PR! I plan to roll this up with a couple of other changes (that I have in progress) to make the next release.

ankithbtik:
Thanks, @rnc & @goldmann. Waiting for the next release... :)

PR #820, opened 2022-11-17
Add type hints.

greenatatlassian:
To ease future development, I've gone ahead and added type hints to the vast majority of the code base. With one exception, all changes are to add type hints only—no code changes.

The exception is that I have made a small change to the code in `descriptor/image.py`, where `self._image_overrides` is now set to a named tuple instead of a dictionary (otherwise I was unable to follow the types through this dictionary).

A couple of call outs:
- I've added a `PathType` which is just an alias for `str`, but hopefully makes it more obvious when we're using a string as a path as opposed to something else. I may not have picked up on all cases where `str` is a path.
- To avoid introducing any circular imports (now or down the road) purely because of type hinting, where the types come from a module that was not already imported, they are protected by `if TYPE_CHECKING` so they are not imported at runtime.
- There are a few places where the code could potentially benefit from structural typing (PEP 544, added Python 3.8). The dependency handler is one such place (though I think we should actually add a `HasDependencies` mixin type).
- Most of the types in `cekit.types` would be better expressed as Type Aliases (PEP 613, added in Python 3.10). It's a few years before we'll be not needing Python 3.9, though. Some of these could also be better addressed by creating domain specific types, rather than using POPOs.
- Regarding types that must be enclosed in quotes: PEP 563 (Python 3.7) would enable us to drop the quotes everywhere by adding a `from __future__` import, however, it looks like instead waiting for PEP 649 would keep the code cleaner, so I have stuck with the original approach.
- The names of descriptors are not (currently) required to be strings, though allowing anything here makes things more complicated. We should consider normalising on strings.
- Maybe I'm missing something, but the osbs builder appears to have some serious type errors. I'm surprised this code works at all (maybe it isn't covered by tests?)
- There are a few places where we replace instance variables with things of different type. So the type of something depends on when you access it. IMHO this is a no-no and very prone to errors (possibly explains what is going on with osbs mentioned above, though if it does, I don't see it.) We should consider refactoring this. Specifically, currently, the things under the `Descriptor` hierarchy serve two purposes: represent the descriptor loaded from disk, and represent the internal view of these things after "hydrating". 
- I have not looked at adding or checking types in the test code at all.

Along the way, I noticed a number of things in the code. I have added TODO comments for those things, and if time allows I may raise a separate PR to address most of those.

### DIFF
--- a/cekit/builder.py
+++ b/cekit/builder.py
@@ -1,8 +1,13 @@
 import logging
+from typing import TYPE_CHECKING, Optional
 
 from cekit.config import Config
 from cekit.errors import CekitError
-from cekit.tools import DependencyHandler
+from cekit.tools import DependencyHandler, Map
+from cekit.types import PathType
+
+if TYPE_CHECKING:
+    from cekit.generator.base import Generator
 
 LOGGER = logging.getLogger("cekit")
 CONFIG = Config()
@@ -12,12 +17,12 @@ class Command(object):
     TYPE_BUILDER = "builder"
     TYPE_TESTER = "tester"
 
-    def __init__(self, command, command_type):
-        self._command = command
-        self._type = command_type
+    def __init__(self, command: str, command_type: str) -> None:
+        self._command: str = command
+        self._type: str = command_type
 
         # Initialize dependency handler
-        self.dependency_handler = DependencyHandler()
+        self.dependency_handler: DependencyHandler = DependencyHandler()
 
         LOGGER.debug("Checking CEKit core dependencies...")
         self.dependency_handler.handle_core_dependencies()
@@ -42,16 +47,16 @@ class Builder(Command):
     Class representing generic builder - if it's instantiated it returns proper builder
     """
 
-    def __init__(self, build_engine, params):
+    def __init__(self, build_engine: str, params: Map) -> None:
         self.params = params
-        self.build_engine = build_engine
+        self.build_engine: str = build_engine
 
-        self.target = self.params.target
-        self.generator = None
+        self.target: PathType = self.params.target
+        self.generator: Optional[Generator] = None
 
         super(Builder, self).__init__(self.build_engine, Command.TYPE_BUILDER)
 
-    def execute(self):
+    def execute(self) -> None:
         self.prepare()
         self.before_generate()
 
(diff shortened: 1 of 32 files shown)


greenatatlassian:
Looks like some simple test failures to fix up still.

Interestingly, I basically cannot run the full CI pipeline locally—I'm on an arm based mac, and a lot of the tests just fall over. If people have ideas, I'd love to be able to raise a PR without failing tests.

greenatatlassian:
Test failures sorted out.

rnc (approved):

rnc:
Firstly, thank you for a such a huge piece of work! That is really excellent especially combined with your diagnosis above. 

I think perhaps some of the points above should be split off into separate tickets so they aren't lost (namely 6/7/8) ? 

I concur its reasonable not to change the tests. As for the PEP comments - those are interesting but we are also trying to keep it compatible with Python versions available on e.g. RHEL. However, this could be discussed separately.

greenatatlassian:
Thanks @rnc. Happy to create separate issues for some/all of those comments—they're really just observations, not something I was hoping to address here. 

Yeah, I figured we should at least be maintaining compatibility with the supported python versions—more notes for down the road when some of these things are available across all versions supported. I learned a lot about the state of Python type hinting!

We may want to think about a way to track handling things that should be updated/cleaned up as we retire older versions. Maybe just something like `TODO: Python 3.7 deprecation...` in the code? Or a single issue where we collect everything that will need to be done when we remove support for a particular python version? Not a big deal, but it would be useful to capture these kinds of things some how so they don't get forgotten.

greenatatlassian:
I've gone ahead and created #822, #823 and #824 to capture those three points @rnc mentions would be worth taking up elsewhere.

rnc:
> We may want to think about a way to track handling things that should be updated/cleaned up as we retire older versions. Maybe just something like `TODO: Python 3.7 deprecation...` in the code? Or a single issue where we collect everything that will need to be done when we remove support for a particular python version? Not a big deal, but it would be useful to capture these kinds of things some how so they don't get forgotten.

That is a good point. I think perhaps `TODO` in the code would be sufficient but I'm open to alternatives.

PR #825, opened 2022-11-24
Issue-821: Enable inclusion/exclusion of features using regex

ankithbtik:
With these changes, CEKit allows users to include or exclude the execution of a single (or multiple) feature file(s) using the regex of the feature files' names.

For example, if there are few integration tests written in the feature files `integ_tests1.feature` and `integ_tests2.feature`, and for a particular flow we want to run only the integration tests, then we can do that by -
`cekit test behave --include-re 'integ_tests'`

Similarly, if we want to exclude the integration tests, then we can do that by -
`cekit test behave --exclude-re 'integ_tests'`

These flags will give more granular control to the users to decide which tests to run and which ones to skip.


cekit/cli.py
```diff
@@ -312,14 +312,16 @@ def test(image, overrides):
     show_default=True,
 )
 @click.option("--wip", help="Run test scenarios tagged with @wip only.", is_flag=True)
+@click.option("--include-re", help="Run only those features which match this regex")
+@click.option("--exclude-re", help="Do not run those features which match this regex")
```
rnc, line 316: Can we add these parameters to the user-guide (in file `docs/handbook/testing/behave.rst` ) please?

rnc (changes requested):

cekit/cli.py
ankithbtik: Documentation added!

ankithbtik:
Hi @rnc , could you please check why https://github.com/cekit/cekit/actions/runs/3592335649/jobs/6047922658 is stuck at 'Setup MacOS' step? Thanks.

tests/test_integ_test_behave.py
```diff
@@ -151,6 +151,86 @@ def test_execute_simple_behave_test_with_overrides(build_image):
...
+
+    os.makedirs(features_dir)
+
+    with open(os.path.join(features_dir, "basic.feature"), "w") as fd:
+        fd.write(feature)
+
+    with Chdir(test_image_dir):
+        result = CliRunner().invoke(
+            cli,
+            ["-v", "test", "behave", "--exclude-re", "basic.feature"],
```
greenatatlassian, line 219: This doesn't really test that the expression is a regular expression. Same with the `include` test above.

rnc (approved):

tests/test_integ_test_behave.py
ankithbtik: It is actually a regex (although it doesn't look like) that tells CEKit to exclude all the feature files which contain 'basic.feature' string in its name. This would have worked the same way if the regex would have been simply 'basic' or 'basic.feature$'.

ankithbtik:
Hi @rnc, thanks for approving the changes. Do you have plans to merge this and release the changes?

rnc:
> Hi @rnc, thanks for approving the changes. Do you have plans to merge this and release the changes?

I've now merged it! I don't have immediate plans for a new release - are you needing one urgently?

ankithbtik:
Thanks! We definitely need it, but, we can wait for some time if you want to club other changes also in the next release. Any estimate around when you are planning for the release will be very helpful.

rnc:
@ankithbtik Hi. I plan to release this week once #831 is completed.

rnc:
@ankithbtik @greenatatlassian 4.5.0 is now released

PR #826, opened 2022-11-25
search podman binary instead hardcode to /usr/bin/podman

spolti:
Signed-off-by: spolti <fspolti@redhat.com>

### DIFF
--- a/.github/workflows/cekit.yml
+++ b/.github/workflows/cekit.yml
@@ -54,6 +54,10 @@ jobs:
       if: startsWith(matrix.os,'macos')
       run: |
         pip install tox
+        brew install podman
+        podman machine init
+        podman machine start
+        podman info
     - name: Run tests
       run: |
         PV=${{ matrix.python-version }}
--- a/cekit/builders/podman.py
+++ b/cekit/builders/podman.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import shutil
 from typing import List
 
 from cekit.builder import Builder
@@ -25,8 +26,12 @@ class PodmanBuilder(Builder):
 
     def run(self):
         """Build container image using podman."""
+
+        podman = shutil.which("podman")
+        if podman is None:
+            raise RuntimeError("podman binary was not found in the system.")
         tags: List[str] = self.params.tags
-        cmd: List[str] = ["/usr/bin/podman", "build"]
+        cmd: List[str] = [podman, "build"]
 
         if not tags:
             tags = self.generator.get_tags()
--- a/tests/test_builder.py
+++ b/tests/test_builder.py
@@ -1,6 +1,7 @@
 import glob
 import logging
 import os
+import shutil
 import subprocess
 import time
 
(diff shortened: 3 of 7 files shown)


cekit/builders/podman.py
```diff
@@ -25,8 +26,12 @@ def dependencies(params=None) -> DependencyDefinition:
 
     def run(self):
         """Build container image using podman."""
+
+        podman = shutil.which("podman")
+        if podman is None:
+            raise RuntimeError("podman binary was not found int he system.")
```
rnc, line 32: Typo : 
```suggestion
            raise RuntimeError("podman binary was not found in the system.")
```

rnc:
Can you also rebase the branch as I've pushed some test fixes.

.github/workflows/cekit.yml
```diff
@@ -54,6 +54,11 @@ jobs:
       if: startsWith(matrix.os,'macos')
       run: |
         pip install tox
+        /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
```
rnc, line 57: According to https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md Homebrew is already installed

rnc:
A number of tests have 
```
@pytest.mark.skipif(
    platform.system() == "Darwin", ....
```

to disable the test if its running Podman. These should be re-enabled I think?

spolti:
> And can you squash the commits please?

will do once I get the tests passing.

spolti:
> to disable the test if its running Podman. These should be re-enabled I think?

yes, I tried to uncomment one to see how it goes, let's see when the GHA starts.

.isort.cfg
```diff
@@ -0,0 +1,3 @@
+[settings]
+# default venv created by intellij
+skip=venv
```
rnc, line 3: This is currently configured in `setup.cfg`. I would rather not add another configuration file.

spolti: oh i see.
thanks.

setup.cfg
```diff
@@ -5,12 +5,13 @@ less-zeroes=True
...
 profile=black
 src_paths=cekit,docs,support,tests
-skip=__init__.py,.tox
+skip=__init__.py,.tox,venv
 
 [flake8]
 max-line-length = 88
 select = C,E,F,W,B,B950
 extend-ignore = E203, E501, W503
+exclude = venv,.tox
```
rnc, line 14: I am not convinced by the `venv` references for flake8 and isort. Its not guaranteed that users will be using `venv` and its often recommended to use `.venv` instead (See https://docs.python.org/3/library/venv.html ) - which is what I also use with https://github.com/MichaelAquilina/zsh-autoswitch-virtualenv/ though I host all my virtualenvs in `$HOME/.virtualenvs`.

cekit/builders/podman.py
spolti: probably I didn't rebase and force pushed at some point.

setup.cfg
spolti: this venv was created by IDE, mine ones are kept on another place.
as you asked before I'll remove it so it can be handled on another PR.

spolti: it should be ready now.

cekit/builders/podman.py
spolti: updated.

PR #827, opened 2022-11-25
Rename cekit.type to avoid IntelliJ import clash. Add more test exclu…

rnc:
…sions.


rnc:
cc @greenatatlassian I've had to rename to avoid problems in IntelliJ (similar to https://youtrack.jetbrains.com/issue/PY-43957/Warn-when-custom-packages-modules-shadow-objects-from-the-standard-library ,  https://youtrack.jetbrains.com/issue/PY-23615) ; not sure if there is a better solution to this.

greenatatlassian:
Good move—I should have realised that wasn't a good name!

PR #838, opened 2023-01-30
Handle upgraded packaging removal of LegacyVersion

### LINKED ISSUE #837: Upgrade CEKit to be compatible with new packaging.
This is a follow on from #830 and due to https://bugzilla.redhat.com/show_bug.cgi?id=2162433 we should now make CEKit fully compatible. This will likely mean removing `LegacyVersion` (etc) references or replacing with suitable alternatives.

rnc:
Fixes #837

### DIFF
--- a/cekit/generator/base.py
+++ b/cekit/generator/base.py
@@ -10,7 +10,7 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional
 from urllib.parse import urlparse
 
 from jinja2 import Environment, FileSystemLoader
-from packaging.version import LegacyVersion
+from packaging.version import InvalidVersion, Version
 from packaging.version import parse as parse_version
 
 from cekit.cekit_types import PathType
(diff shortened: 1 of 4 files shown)


cekit/generator/base.py
```diff
@@ -723,21 +724,25 @@ def add_module(self, module: Module):
...
-                (
-                    "Module's '{}' version '{}' does not follow PEP 440 versioning scheme "
-                    "(https://www.python.org/dev/peps/pep-0440), "
-                    "we suggest follow this versioning scheme in modules"
-                ).format(module.name, version)
+        default_version: Version = parse_version(self._defaults.get(module.name))
+        try:
+            current_version: Version = parse_version(version)
+            # https://github.com/pypa/packaging/issues/631#issuecomment-1343508993
+            if not hasattr(current_version, "major"):
```
rnc, line 731: This will only happen under Python 3.6 if packaging is < 22 and returns a LegacyVersion. At this point we force it into the invalid version path to handle LegacyVersions.

goldmann: We should drop support for Python 3.6 since it is EOL.

rnc: Dropping Python 3.6 could be considered but I think that should be handled separately.

goldmann (approved):

PR #850, opened 2023-05-17
Adjust env schema to permit extra keys

### LINKED ISSUE #846: Add new fields to module.yaml envs
We are doing some automation that uses module.yaml. 
We would need need fields added to envs. Currently we only have name and description, we need required="true"|"false" and usage="build-time"|"execution-time".

rnc:
Fixes #846

### DIFF
--- a/cekit/descriptor/env.py
+++ b/cekit/descriptor/env.py
@@ -12,8 +12,7 @@ mapping:
   name: {type: str, required: True}
   value: {type: any}
   example: {type: any}
-  description: {type: str}
-  metadata: {type: any}"""
+  description: {type: str}"""
 )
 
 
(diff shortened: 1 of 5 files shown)


cekit/descriptor/env.py
```diff
@@ -6,11 +6,14 @@
...
-map:
+type: map
+allowempty: True
+mapping:
   name: {type: str, required: True}
   value: {type: any}
   example: {type: any}
-  description: {type: str}"""
+  description: {type: str}
+  metadata: {type: any}"""
```
goldmann, line 16: So, in the end we add the `metadata` key as well? I thought we dropped the idea.

rnc: I guess @jmtd missed that! Let me remove it...

goldmann: Sure, let's blame @jmtd! :)

goldmann (approved):

PR #857, opened 2023-05-26
Improve exception handling and Python 36 subprocess compatibility

### LINKED ISSUE #856: Python 3.6 and get_brew_url
@spolti reported that under Python 3.6 rhpkg/brew was not correctly being called. This was traced to get_brew_url which was using a subprocess.run using code > python 3.6 and the CI job was running on RHEL under Python 3.6

The solution is to change to use the run_wrapper which handles the differences.

rnc:
Fixes #856 

@spolti FYI

### DIFF
--- a/cekit/builders/docker_builder.py
+++ b/cekit/builders/docker_builder.py
@@ -110,7 +110,7 @@ class DockerBuilder(Builder):
                 if layer_id_match:
                     docker_layer_ids.append(layer_id_match.group(1))
 
-        except requests.ConnectionError as ex:
+        except requests.ConnectionError:
             exception_chain = traceback.format_exc()
             LOGGER.debug(
                 "Caught ConnectionError attempting to communicate with Docker ",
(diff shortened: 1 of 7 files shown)


cekit/builders/docker_builder.py
```diff
@@ -150,7 +147,7 @@ def _build_with_docker(self, docker_client):
                     "definitions."
                 )
 
-            raise CekitError(msg, ex)
+            raise CekitError(msg) from ex
```
goldmann, line 150: Didn't know about this construct!

goldmann (dismissed):

cekit/builders/docker_builder.py
rnc: Realised the CEKitErrors weren't chaining the exceptions so looked up https://peps.python.org/pep-3134/ and https://blog.ram.rachum.com/post/621791438475296768/improving-python-exception-chaining-with

spolti (approved):

goldmann (approved):

PR #866, opened 2023-09-06
Switch from git to https protocol

rnc:


rnc:
> I wondered if cekit would need to be updated for this. My dist-git clones (created via cekit although a long time ago) are git+ssh (not plain git), and are still working, FWIW.

See internal CEKit channel - I've cc'd you showing a backtrace.

PR #868, opened 2023-09-06
Add support for ARG

### LINKED ISSUE #865: Provide support to the ARG property
Sometimes there is a need to have a build argument to be used in a multi-stage build, as part o f the build process.

In this case, we would like to have CeKit support the usage of the ARG property in the `image.yaml` descriptor.
A simple use case can be found in this Dockerfile:

```
FROM quay.io/kiegroup/kogito-swf-builder-nightly:latest AS builder

ARG QUARKUS_EXTENSIONS

# Copy from build context to skeleton resources project    
COPY --chmod=777 * ./resources/

RUN /home/kogito/launch/build-app.sh ./resources

RUN ls -lah
RUN pwd
RUN cat /home/kogito/serverless-workflow-project/pom.xml
RUN ls -lah  /home/kogito/serverless-workflow-project/target/
#=============================
# Runtime Run
#=============================
FROM registry.access.redhat.com/ubi8/openjdk-11:latest

ENV LANG='en_US.UTF-8' LANGUAGE='en_US:en'

# We make four distinct layers so if there are application changes the library layers can be re-used
COPY --from=builder --chown=185 /home/kogito/serverless-workflow-project/target/quarkus-app/lib/ /deployments/lib/
COPY --from=builder --chown=185 /home/kogito/serverless-workflow-project/target/quarkus-app/*.jar /deployments/
COPY --from=builder --chown=185 /home/kogito/serverless-workflow-project/target/quarkus-app/app/ /deployments/app/
COPY --from=builder --chown=185 /home/kogito/serverless-workflow-project/target/quarkus-app/quarkus/ /deployments/quarkus/

EXPOSE 8080
USER 185
ENV AB_JOLOKIA_OFF=""
ENV JAVA_OPTS="-Dquarkus.http.host=0.0.0.0 -Djava.util.logging.manager=org.jboss.logmanager.LogManager"
ENV JAVA_APP_JAR="/deployments/quarkus-run.jar"
```

**Additional context**
Add any other context about the feature request here.

rnc:
Fixes #865

### DIFF
--- a/docs/descriptor/includes/args.rst
+++ b/docs/descriptor/includes/args.rst
@@ -1,0 +1,21 @@
+Args
+------
+
+Key
+    ``args``
+Required
+    No
+
+Every image can include arg instructions. See https://docs.docker.com/engine/reference/builder/#arg for more
+details. CEKit makes it easy to do so with the ``args`` section. An optional default value may be provided.
+
+.. note::
+
+    ARGS will be rendered before any environment variables, labels or execution sections.
+
+.. code-block:: yaml
+
+    args:
+        - name: "user1"
+          value: "someuser"
+        - name: "QUARKUS_EXTENSION"
--- a/docs/descriptor/includes/module/execute.rst
+++ b/docs/descriptor/includes/module/execute.rst
@@ -22,3 +22,12 @@ as defined.
         # Then the redefine.sh file will be executed as jboss user
         - script: redefine.sh
           user: jboss
+
+This will be rendered as
+
+.. code-block:: sh
+
+    RUN [ "sh" "-x" "/tmp/scripts/install.sh" ]
+
+    USER jboss
+    RUN [ "sh" "-x" "/tmp/scripts/redefine.sh" ]
--- a/tests/test_dockerfile.py
+++ b/tests/test_dockerfile.py
@@ -776,12 +776,37 @@ def test_package_reinstall(tmpdir):
     regex_dockerfile(target, "reinstall -y tzdata")
 
 
-def test_args(tmpdir):
+def test_args_podman(tmpdir):
     target = str(tmpdir.mkdir("target"))
 
     generate(
         target,
         ["-v", "build", "--dry-run", "--container-file", "Dockerfile", "podman"],
+        descriptor=[
+            {
+                "release": 1,
+                "version": 1,
+                "from": "fromimage",
+                "name": "targetimage",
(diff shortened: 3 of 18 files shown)


docs/descriptor/includes/args.rst
```diff
@@ -0,0 +1,17 @@
...
+    ``args``
+Required
+    No
+
+Every image can include arg instructions. See https://docs.docker.com/engine/reference/builder/#arg for more
+details. CEKit makes it easy to do so with the ``args`` section. An optional default value may be provided.
+
+.. code-block:: yaml
+
+    labels:
```
goldmann, line 14: This example refers to labels, not args.

tests/test_dockerfile.py
```diff
@@ -776,6 +776,37 @@ def test_package_reinstall(tmpdir):
...
+        target,
+        ["-v", "build", "--dry-run", "--container-file", "Dockerfile", "podman"],
+        descriptor=[
+            {
+                "release": 1,
+                "version": 1,
+                "from": "fromimage",
+                "name": "targetimage",
+                "args": [{"name": "foo", "value": "bar"}, {"name": "labela"}],
+                "labels": [
```
goldmann, line 792: Can we remove labels from this example? It's not relevant and distracts a bit.

goldmann (changes requested):
Generally, I think this is the first step towards `ARGS`  support: we can provide and these will be placed in the Dockerfile. The second step, and this is what is missing, is the ability to populate these at the build time.

Other question I have is that shouldn't args be part of the module? Because defining them without using them is not very useful :) I guess it could be useful at the image level as well (defining global labels?).

So, to fully implement it we probably need:

1. Some example of a proper usage + test + extended docs
2. Ability to populate it at build time

docs/descriptor/includes/module/execute.rst
```diff
@@ -21,4 +21,13 @@ as defined.
...
         - script: redefine.sh
-          user: jboss
\ No newline at end of file
+          user: jboss
+
+This will be rendered as
+
+.. code-block:: sh
+
+    RUN [ "sh" "-x" "/tmp/scripts/install.sh" ]
```
goldmann, line 30: Copy/paste error?

goldmann (approved):

docs/descriptor/includes/module/execute.rst
goldmann: My bad, all good.

Issue #886, opened 2024-01-30
Opt-out from package manager cleanup 


ricardozanini:
**Describe the solution you'd like**
Hi! Is it possible to opt-out of the package manager metadata clean up? I can manage this by myself in a custom module.

**Additional context**
I'm trying to produce [reproducible builds for my images](https://reproducible-builds.org/) and this package manager cleanup that Cekit does at the end:

```Dockerfile
    # Clear package manager metadata
    RUN microdnf clean all && [ ! -d /var/cache/yum ] || rm -rf /var/cache/yum
```

Adds the `rpm` and `microdnf` cache and history I cleared before. These files contain non-deterministic files with timestamps and random UUID, making a reproducible build impossible.

I'm also opened to suggestions and how to approach this.

rnc:
Related to this is https://github.com/cekit/cekit/issues/849 where they wanted package manager cleanup after each install in each layer. 

There is also the point that I suspect `/var/cache/yum` might be outdated now - it should be (or be as well) `/var/cache/dnf`. And the `if` test is somewhat redundant with `rm -rf`. 

In terms of package manager configuration - perhaps it could be more flexible offering the current choice (i.e. cleanup at end), cleanup per module, or no cleanup ?

ricardozanini:
I'd vote for no cleanup at all and let the module or image handle that. I was maintaining some old codebase that was running the clean up in a "clean up" module. :) 

I understand that the tool shouldn't be opinionated to this point.

rnc:
I don't think no cleanup is an option - that would be a rather surprising change to many users and effectively a 'breaking' change. However, I understand some users want more flexibility and manual control hence looking at configuration instead. 

@goldmann - thoughts?

goldmann:
I'm with you, @rnc. I think cleaning up by default is a sane default. If it would be possible to not make it as part of a switch, then that would be nice. But I imagine this is all-or-nothing option everything cleaned up or nothing cleaned up and left to module developer to tackle this.

ricardozanini:
+1 to add the on/off option for clean up.

jmtd:
I think the package management abstraction in `template.jinja` needs a little clean-up. At least with `manager: microdnf`, there's a `clean all` stage added to the end of each `install` as well. Thinking about this from a few different perspectives

 * I think supporting reproducible builds is a very good idea
 * if using layer squashing, from a speed perspective, doing all those `clean`s except the last are just costing build time
[...]

PR #888, opened 2024-02-07
template.jinja: also clean up /var/lib/{rpm,dnf} for reproducibility

jmtd:
hopefully addresses #886 .

### DIFF
--- a/cekit/templates/template.jinja
+++ b/cekit/templates/template.jinja
@@ -74,7 +74,7 @@ rm{% for repo in repositories %} /etc/yum.repos.d/{{ repo.filename }}{% endfor %
  #}
 {%- macro repo_clear_cache(pkg_manager) -%}
     {% if pkg_manager in helper.SUPPORTED_PACKAGE_MANAGERS %}
-rm -rf "/var/cache/yum" "/var/lib/rpm" "/var/lib/dnf" "/var/cache/apt"
+rm -rf "/var/cache/yum" "/var/lib/rpm" "/var/lib/dnf" "/var/cache/apt" "/var/cache/dnf"
     {% endif %}
 {%- endmacro -%}
 
--- a/cekit/templates/template.jinja
+++ b/cekit/templates/template.jinja
@@ -74,7 +74,7 @@ rm{% for repo in repositories %} /etc/yum.repos.d/{{ repo.filename }}{% endfor %
  #}
 {%- macro repo_clear_cache(pkg_manager) -%}
     {% if pkg_manager in helper.SUPPORTED_PACKAGE_MANAGERS %}
-{{ pkg_manager }} clean all && [ ! -d /var/cache/yum ] || rm -rf /var/cache/yum
+rm -rf "/var/cache/yum" "/var/lib/rpm" "/var/lib/dnf"
     {% endif %}
 {%- endmacro -%}
 
@@ -266,8 +266,7 @@ rm{% for repo in repositories %} /etc/yum.repos.d/{{ repo.filename }}{% endfor %
     {% if image.from != 'scratch' %}
     # Switch to 'root' user and remove artifacts and modules
     USER root
-    RUN [ ! -d /tmp/scripts ] || rm -rf /tmp/scripts
-    RUN [ ! -d /tmp/artifacts ] || rm -rf /tmp/artifacts
+    RUN rm -rf "/tmp/scripts" "/tmp/artifacts"
     {% endif -%}
 
     {% if helper.packages_to_install(image) or helper.packages_to_remove(image) or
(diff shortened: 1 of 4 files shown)


jmtd:
I have manually verified that a downstream image can do e.g. `microdnf install...` without problems. I should probably add some tests to this PR.

cekit/templates/template.jinja
```diff
@@ -74,7 +74,8 @@ rm{% for repo in repositories %} /etc/yum.repos.d/{{ repo.filename }}{% endfor %
  #}
 {%- macro repo_clear_cache(pkg_manager) -%}
     {% if pkg_manager in helper.SUPPORTED_PACKAGE_MANAGERS %}
-{{ pkg_manager }} clean all && [ ! -d /var/cache/yum ] || rm -rf /var/cache/yum
+{{ pkg_manager }} clean all \
```
rnc, line 77: Do we really want `clean all` still here?

cekit/templates/template.jinja
```diff
@@ -74,7 +74,8 @@ rm{% for repo in repositories %} /etc/yum.repos.d/{{ repo.filename }}{% endfor %
...
+        && rm -rf "/var/cache/yum" "/var/lib/rpm" "/var/lib/dnf"
```
rnc, line 78: What about `/var/cache/apt` ? And `/var/cache/dnf` ?

tests/test_dockerfile.py
```diff
@@ -825,6 +825,25 @@ def test_args_buildah(tmpdir):
...
+                "manager": "microdnf",
+                "repositories": [{"name": "foo", "rpm": "foo-repo.rpm"}],
+                "install": ["a"],
+            },
+            "osbs": {"repository": {"name": "repo_name", "branch": "branch_name"}},
+        },
+    )
+    regex_dockerfile(target, "rm -rf.*/var/cache/yum")
+    regex_dockerfile(target, "rm -rf.*/var/lib/rpm")
+    regex_dockerfile(target, "rm -rf.*/var/lib/dnf")
```
rnc, line 847: I think you need to run `./support/run_formatter.py` to correct the formatting.

cekit/templates/template.jinja
jmtd: I was going to try and avoid Debian stuff for this PR (as there are lots more things to fix there and I'm thinking of working on a larger refactoring of the pkg manager) but I could add the relevant paths here if you wish.

cekit/templates/template.jinja
jmtd: No, I'll drop it. Thanks

cekit/templates/template.jinja
jmtd: microdnf didn't write to `/var/cache/dnf. Perhaps full-fat `dnf` does? I'll check

rnc: If we are removing clean all then we have to add the apt packages to remove as otherwise this PR will break apt clean.

jmtd: > microdnf didn't write to `/var/cache/dnf`. Perhaps full-fat `dnf` does? I'll check

full DNF does so I'll add that to the list

jmtd: > If we are removing clean all then we have to add the apt packages to remove as otherwise this PR will break apt clean.

Ack, addressed in [9fe526b](https://github.com/cekit/cekit/pull/888/commits/9fe526bddd436241ac49169403774bb4b8b5cbbe).

jmtd: > > microdnf didn't write to `/var/cache/dnf`. Perhaps full-fat `dnf` does? I'll check
> 
> full DNF does so I'll add that to the list

Addressed in [3681255](https://github.com/cekit/cekit/pull/888/commits/36812557cbc98ad31d45d7870e36ed8910ddda70).

jmtd:
Further test failures to adjust

`tests/test_integ_builder_osbs.py::test_osbs_builder_with_rhpam_1`: the test has an _entire rendered Dockerfile_ embedded in it and checks for it byte-for-byte identical! This is going to break for any change we make to the template. I think that test (and any more like it) need refactoring to limit their comparisons to the topic of the test.

rnc:
> Further test failures to adjust
> 
> `tests/test_integ_builder_osbs.py::test_osbs_builder_with_rhpam_1`: the test has an _entire rendered Dockerfile_ embedded in it and checks for it byte-for-byte identical! This is going to break for any change we make to the template. I think that test (and any more like it) need refactoring to limit their comparisons to the topic of the test.

I think at the time I did that to make sure the multistage was accurately building what the users wanted. While fiddly it was meant to catch mistakes in the generation.

rnc (approved):

PR #893, opened 2024-02-28
Only require odcs client when using content_sets

### LINKED ISSUE #702: Dependency Handling with descriptor reference
This is a sequel to #696 / #701 ; if we instead we modify builder.py :: before_generate (L81) so that init is called _before_ `dependency_handler.handle` then we could pass the descriptor reference to the static `dependencies` method thereby allowing it to check for content sets and determine whether to require the odcs dependency.

Notes from discussion: External code https://github.com/cekit/behave-test-steps/blob/v1/loader.py requires similar parameters so we can't change the number of parameters. It might be possible to overload the params, perhaps using `params.image....repo... `and check for content_sets.

rnc:
Fixes #702 #696

### DIFF
--- a/cekit/tools.py
+++ b/cekit/tools.py
@@ -647,18 +647,11 @@ class DependencyHandler(object):
         if not o:
             return
 
-        # Get the class of the object
-        clazz = type(o)
+        if callable(getattr(o, "dependencies", None)):
+            from cekit.generator.base import Generator
 
-        for var in [clazz, o]:
-            # Check if a static method or variable 'dependencies' exists
-            dependencies = getattr(var, "dependencies", None)
-
-            if not dependencies:
-                continue
-
-            # Check if we have a method
-            if callable(dependencies):
-                # Execute that method to get list of dependencies and try to handle them
+            # Execute that method to get list of dependencies and try to handle them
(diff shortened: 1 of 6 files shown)


cekit/tools.py
```diff
@@ -647,18 +647,11 @@ def handle(self, o: Any, params: Map) -> None:
         if not o:
             return
 
-        # Get the class of the object
-        clazz = type(o)
+        if callable(getattr(o, "dependencies", None)):
```
goldmann, line 650: I think this should be fine. I briefly checked and we don't have any static methods on which we would need to execute it.

jmtd (approved):
LGTM!
Review notes: built into a virtualenv, odcs library was not installed; build of openjdk container image (w/o using content sets) worked; build with content-sets turned on failed (as expected) with helpful error message.

goldmann (approved):

cekit/tools.py
rnc: It does actually call e.g.:
```
    @staticmethod
    def dependencies(params: Map = None, image: Image = None) -> DependencyDefinition:
```

but brief testing seems to imply getattr works for static and non-static

PR #897, opened 2024-03-06
Update sphinx and sphinx-autobuild. Switch theme to Furo.

### LINKED ISSUE #896: Update Documentation Theme
Currently we lock the sphinx-autodocs and prevent sphinx from being upgraded. The early autodocs is now hard to setup locally due to its age. The guzzle theme is now unmaintained and incompatible with later sphinx versions. We should change from guzzle to another theme listed on e.g. https://sphinx-themes.org/#themes

On several of the themes below I have found it hard to display the caption `Container image creation tool` on the sidebar. Most themes now provide their own sidebars i.e. the template customisation, the CSS etc all may not be required.

# [Guzzle](https://github.com/guzzle/guzzle_sphinx_theme/)
This is the current theme and is rendered https://docs.cekit.io/en/develop/index.html 

It does support breadcrumbs. There is no right hand side contents. Inline contents work fine. Navigation does float when scrolling. Search does provide context.

# [Sphinx-Read-The-Docs](https://sphinx-rtd-theme.readthedocs.io/en/stable/) 
image

It does support breadcrumbs. There is no right hand side contents. Inline contents work fine. Navigation does float when scrolling. Search does provide context.

# [Furo](https://pradyunsg.me/furo/)
image

It doesn't currently support breadcrumbs. Inline contents are not required and either need to have a special class added to them to allow them or removed (and rely on the righthand sidebar). Supports dark/light theme. Navigation does float when scrolling. Search does provide context.

# [Alabastor](https://alabaster.readthedocs.io/en/latest/)
image

Very minimal, similar to current theme.  No breadcrumbs or right hand sidebar. Does support `description` tag to add blurb under logo. It is not possible to display the CEKit version from what I can see. Navigation does not float when scrolling.

# [SphinxBook Theme](https://sphinx-book-theme.readthedocs.io/en/latest/reference.html)
image
Similar to Furo. Looks like it derives from https://pydata-sphinx-theme.readthedocs.io. Doesn't support breadcrumbs. Has a righthand sidebar (which effectively duplicates inline contents like Furo). Navigation does float when scrolling. Supports extra buttons like full-screen, dark/light mode. Doesn't have a footer containing links to Sphinx etc. Search does not provide context.

# [SphinxAwesome](https://sphinxawesome.xyz)
image
Interesting configuration system using a proper Python object. Does support breadcrumbs. Highlighting better than most. Light/dark mode. Search has context. Annoying paragraph indicator on each heading. Screen 'flash' on every page change which is a known issue but makes using this impossible.

rnc:
Fixes #896


jmtd:
On pages such as descriptor/image.html , the ToC on the RHS is much better than the current site approach

jmtd:
dark mode / light mode switching works. dark theme looks good

jmtd:
the diagrams work on this / the example build where they are broken on the current docs site

jmtd:
Not a blocker: diagrams don't respond to dark mode
darkmode

jmtd (approved):
Thanks for doing this!

rnc:
> the diagrams work on this / the example build where they are broken on the current docs site

The diagrams do work on devel : this change fixed it https://github.com/cekit/cekit/pull/892

> Not a blocker: diagrams don't respond to dark mode ![darkmode](https://private-user-images.githubusercontent.com/63694/310831972-f9b2617d-d66e-4b69-b394-ffc4498d1a97.png?

I think all the diagraph generated PNGs would need to be either redone with transparent backgrounds or alternates supplied for the dark mode. It looks like its possible for some ; I'll take a look.

PR #902, opened 2024-03-12
Remove support for older docker library. 

rnc:
The old library dates from the days it was called "docker-py", the last release (1.10.6) for which was November 2016 (See https://docker-py.readthedocs.io/en/stable/change-log.html#id156 ) 

Originally I considered locking the version to the _current_ version of docker python library (`constants.DEFAULT_DOCKER_API_VERSION`) but it may move too quick and therefore we switched to using our own constant that the behave suite can also reuse.

### DIFF
--- a/cekit/builders/docker_builder.py
+++ b/cekit/builders/docker_builder.py
@@ -32,6 +32,7 @@ except ImportError:
     pass
 
 ANSI_ESCAPE = re.compile(r"\x1B\[[0-?]*[ -/]*[@-~]")
+DOCKER_API_VERSION = "1.35"
 
 
 class DockerBuilder(Builder):
@@ -194,7 +195,7 @@ class DockerBuilder(Builder):
                 )
             )
 
-        params = {"version": docker.constants.DEFAULT_DOCKER_API_VERSION}
+        params = {"version": DOCKER_API_VERSION}
         params.update(docker.utils.kwargs_from_env())
         params["timeout"] = timeout
 
--- a/cekit/builders/docker_builder.py
+++ b/cekit/builders/docker_builder.py
@@ -20,7 +20,6 @@ except ImportError:
 try:
     # Docker Python library, the new one
     import docker
-    from docker.api.client import APIClient as APIClientClass
 except ImportError:
     pass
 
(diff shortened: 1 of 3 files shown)


cekit/builders/docker_builder.py
```diff
@@ -202,12 +194,12 @@ def _docker_client(self):
                 )
             )
 
-        params = {"version": "1.35"}
+        params = {"version": docker.constants.DEFAULT_DOCKER_API_VERSION}
```
rnc, line 197: Currently 1.41 looking at https://github.com/docker/docker-py/blob/7.0.0/docker/constants.py#L4

jmtd (approved):
👍🏼

cekit/builders/docker_builder.py
```diff
@@ -32,14 +31,8 @@
...
 
-try:
-    # Docker Python library, the old one
-    import docker  # noqa: F811
-    from docker.client import Client as APIClientClass  # noqa: F811
-except ImportError:
-    pass
-
 ANSI_ESCAPE = re.compile(r"\x1B\[[0-?]*[ -/]*[@-~]")
+DOCKER_API_VERSION = "1.35"
```
goldmann, line 35: Is this the current API version? Do we have any knowledge about supported API versions in the current library?

What I'm thinking is: we should not make it recent if the library supports older versions as well. The reason is that not all packages/systems around the world are up to date. Another thing is that we probably don't use any features from new APIs.

rnc: We last changed it in 2002 responding to https://github.com/cekit/cekit/pull/774 when the original version (like it **currently** is in Behave (cc @jmtd)) was too old. The compatibility matrix is https://docs.docker.com/engine/api/#api-version-matrix ; unfortunately they don't give dates but the next API change (1.36) came with 18.02 and that was 6 years ago. So IMHO we're still picking a rather old version. The API version I nearly picked (1.41) came from 20.10 and is 3 years ago.

rnc: @jmtd @goldmann  Do we know what versions of Docker Engine are available for RHEL, Debian, etc? 
GHActions use 1.43 API I think from looking at a recent run

goldmann (approved):

cekit/builders/docker_builder.py
goldmann: > @jmtd @goldmann Do we know what versions of Docker Engine are available for RHEL, Debian, etc? GHActions use 1.43 API I think from looking at a recent run

Technically, there is no Docker in RHEL/CentOS Stream. So I imagine the Docker's `docker-engine` packages provide latest versions.

Issue #905, opened 2024-04-03
post-build cleanup of /var/lib/rpm breaks rpm in resulting images


jmtd:
**Describe the bug**
Because we remove `/var/lib/rpm` in clean-up (8873315469583f82fd3705d94b935f2eb494d789 , this was me) , `rpm` does not function properly in downstream images

**To reproduce**
build a redhat or fedora-based container with cekit 4.11.0 or newer.
try to run rpm in the result

```
$ docker run --rm -ti 233a9acf2623 rpm -qa java-*
error: cannot open Packages database in /var/lib/rpm
```

**Expected behavior**

e.g. 

```
$ docker run --rm -ti ubi9/openjdk-17:1.18 rpm -qa java-*
java-17-openjdk-headless-17.0.10.0.7-2.el9.x86_64
java-17-openjdk-17.0.10.0.7-2.el9.x86_64
java-17-openjdk-devel-17.0.10.0.7-2.el9.x86_64
java-17-openjdk-jmods-17.0.10.0.7-2.el9.x86_64
```

rnc:
I think this was expected given the discussion on https://github.com/cekit/cekit/issues/886 ?

jmtd:
I somehow missed that `rpm` would cease functioning. I think that's a big problem.

rnc:
@ricardozanini As you raised #886 I think you should be aware of this. 
@jmtd If that directory is being left, this means the sqllite DB is then still there. Does this then mean we need back the microdnf clean commands?

ricardozanini:
> I somehow missed that `rpm` would cease functioning. I think that's a big problem.

yeah, once we remove the directories `rpm` won't work anymore. The image is supposed to be final based on the reproducible builds approach. Although, it can potentially break new layers of using rpm.

ricardozanini:
@jmtd I'm not using Cekit package manager, but relying on base images that already have what I need. For the other libraries, I'm using tar/untar, manual scripting to install what I need. It's a pain, but at least we can fit into the reproducible build patterns.

I'd say that you can safely revert, but I think we should think in a way to support reproducible builds in Cekit in a different issue/requirements.

rnc:
[...]

PR #906, opened 2024-04-03
Don't remove /var/lib/rpm (#905)

jmtd:
The post-install clean-up is too zealous, and breaks the use of rpm in the resulting images.

Fixes: #905.

### DIFF
--- a/cekit/templates/template.jinja
+++ b/cekit/templates/template.jinja
@@ -74,7 +74,7 @@ rm{% for repo in repositories %} /etc/yum.repos.d/{{ repo.filename }}{% endfor %
  #}
 {%- macro repo_clear_cache(pkg_manager) -%}
     {% if pkg_manager in helper.SUPPORTED_PACKAGE_MANAGERS %}
-rm -rf "/var/cache/yum" "/var/lib/rpm" "/var/lib/dnf" "/var/cache/apt" "/var/cache/dnf"
+rm -rf "/var/cache/yum" "/var/lib/dnf" "/var/cache/apt" "/var/cache/dnf"
     {% endif %}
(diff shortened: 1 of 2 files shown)


rnc:
@jmtd You'll need to rebase to pick up the MacOS fixes from #909

cekit/templates/template.jinja
```diff
@@ -74,7 +74,7 @@ rm{% for repo in repositories %} /etc/yum.repos.d/{{ repo.filename }}{% endfor %
  #}
 {%- macro repo_clear_cache(pkg_manager) -%}
     {% if pkg_manager in helper.SUPPORTED_PACKAGE_MANAGERS %}
-rm -rf "/var/cache/yum" "/var/lib/rpm" "/var/lib/dnf" "/var/cache/apt" "/var/cache/dnf"
+rm -rf "/var/cache/yum" "/var/lib/dnf" "/var/cache/apt" "/var/cache/dnf"
```
rnc, line 77: If we're not removing the var/lib/rpm do we need the microdnf clean ? Will that shrink the DB ?

jmtd: TBD -- I need to dig a bit deeper.

jmtd: No - in fact running it makes the image fractionally larger. The `rm`s do the equivalent job.

jmtd:
Hi @rnc :
> @jmtd You'll need to rebase to pick up the MacOS fixes from https://github.com/cekit/cekit/pull/909

Done, thanks.

I've reviewed whether `microdnf clean -y` would add any value and I don't think it would.

This is ready for another look now please!

rnc (approved):

PR #910, opened 2024-04-03
Remove old Jenkins/CI tag skip

rnc:


jmtd (changes requested):
I support getting rid of this. I think it may be worth a note somewhere that it's gone, release notes if we have them! There's also a  docs reference that might need to go with it:

```
docs/handbook/testing/behave.rst:    If CEKit is **not** running as a user called ``jenkins``, the tag ``@ci``
```

jmtd (approved):

Issue #915, opened 2024-05-28
Ability to inject Args before module processing


etsauer:
**Describe the solution you'd like**
I would like the ability to define `args` at the image descriptor and override level that that get defined in the Containerfile before the module sections get built out. 

A common practice for me (pre CEkit) is to use ARG statements to dynamically set certain package names or versions before installing packages, and have them consumed as follows:

```
ARG OPENSSL_PKG="compat-openssl11"

...

RUN dnf -install -y ${OPENSSL_PKG} ...
```

This allows me to more easily change package names when I switch FROM images between slightly different builds, like ubi ro ubi-minimal or ubi to fedora, where certain package names can differ slightly. 

When I try defining an arg in my image descriptor or in an override like so:

```
###### START module 'developer-base:17'

RUN dnf -install -y ${OPENSSL_PKG} ...

###### END module 'developer-base:17'

###### START image 'dev-tools/dev-workspace-nested-fedora:fedora39'

        ARG OPENSSL_PKG="compat-openssl11"

###### END image 'dev-tools/dev-workspace-nested-fedora:fedora39'
```

I'm sure there are good reasons for processing the image descriptor last, but it would be really great to have some sort of option to inject some sort of image level args that can be parsed by modules.

rnc:
Fixed by #928

PR #928, opened 2024-11-09
Fix overrides for args. Allow args to be injected first for the prima…

rnc:
…ry image

[ Experiment for #915 - awaiting users opinions. The overrides for args fix is required regardless ]

### DIFF
--- a/cekit/descriptor/image.py
+++ b/cekit/descriptor/image.py
@@ -83,6 +83,7 @@ class Image(Descriptor):
         self._descriptor["labels"] = [
             Label(x) for x in self._descriptor.get("labels", [])
         ]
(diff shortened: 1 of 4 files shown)


jmtd:
I think ARG would not be the only exception: `packages.manager` is also interpreted before modules I think.

rnc:
> I think ARG would not be the only exception: `packages.manager` is also interpreted before modules I think.

I think that is already reflected in https://docs.cekit.io/en/latest/handbook/modules/image_order.html as repository management is listed before modules there?

jmtd:
Yes, thanks — sorry my comment was out of context. I was weighing up the pros and cons of this change. I think doing ARGS in the image descriptor before modules is more intuitive, but having that be exceptional compared to other keys makes the model more complicated. Then I realised that it would not be the only exception.

PR #932, opened 2025-02-18
Simplify template.jinja and helper

### LINKED ISSUE #889: refactor template.jinja to fully abstract package manager specifics
Spinning this out of #886: something I've been thinking of a while, I think we should have all package manager specifics out of `template.jinja` and handled elsewhere, perhaps via `TemplateHelper`. For example, stuff like 

```
{%- macro pkg_install(pkg_manager, pkg_manager_flags, packages) -%}
{{ pkg_manager }} {{ helper.package_manager_flags(pkg_manager, pkg_manager_flags) }} {{ helper.package_manager_install(pkg_manager) }} {%- for package in packages %} {{ package }}{% endfor %} \
    {% if pkg_manager == 'microdnf' %}
            && {{ pkg_manager }} clean all \
...
 ```

it's awkward to have both a set-test (`in SUPPORTED_PACKAGE_MANAGERS`) and branching logic on hard-coded specific values (`if pkg_manager == 'microdnf'`) and these macros are harder to read, write or test compared to normal Python code.

I think we should have some kind of abstract class encapsulating all package manager stuff with specific sub-classes for supported package managers, all in proper python code. Something like 

```
class PackageManager(object):
  def update_metadata(self): raise NotImplementedError(...)
  def install_packages(self, pkgs): raise NotImplementedError(...)
  ...
...
class Microdnf(PackageManager):
...
```

and the templates could simplify a lot down to something like `helper.get_manager().install_packages(packages)`.

This would also make adding new package managers much easier.

rnc:
Fixes #889

### DIFF
--- a/cekit/template_helper.py
+++ b/cekit/template_helper.py
@@ -2,7 +2,6 @@ import os
 
 
 class TemplateHelper(object):
-    SUPPORTED_PACKAGE_MANAGERS = ["yum", "dnf", "microdnf", "apk", "apt-get"]
 
     def __init__(self, module_registry):
         self._module_registry = module_registry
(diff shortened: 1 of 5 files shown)


jmtd (approved):
LGTM - ran it on my current test image, all looks good

goldmann (approved):

PR #937, opened 2025-04-29
Add ability to specify example/description for args

### LINKED ISSUE #936: Suport description and example in Args
Similar to env, args must support description or example params for documentation.

rnc:
Fixes #936


jmtd:
Before this PR (testing with cekit 4.13.0): adding description and example keys to `args:`  does basically nothing -- in particular cekit does not complain about invalid keys. After this patch: same behaviour! The ARG is placed earlier in the generated Dockerfile, but i don't know if that's due to a separate change that occurred since 4.13.0.

Testing continues…

jmtd (approved):

PR #939, opened 2025-04-30
Support help template in modules

### LINKED ISSUE #935: help doesn´t work inside module
I like to modify help template and share it with multiples images like a module, but help inside a module is ignored. Can I share help template between multiples repositories?

.

rnc:
Fixes #935

### DIFF
--- a/cekit/descriptor/image.py
+++ b/cekit/descriptor/image.py
@@ -533,8 +533,11 @@ class Image(Descriptor):
                 if self.help.get("template", "") and not os.path.isabs(
                     module.help.get("template")
                 ):
-                    self.help["template"] = "target/image/modules/" + self.help.get(
-                        "template"
+                    self.help["template"] = (
+                        "target/image/modules/"
+                        + to_install.name
+                        + "/"
+                        + self.help.get("template")
                     )
 
             # move this module to the end of the list.
--- a/cekit/descriptor/image.py
+++ b/cekit/descriptor/image.py
@@ -1,5 +1,6 @@
 import copy
 import logging
+import os
 from collections import OrderedDict
 from typing import TYPE_CHECKING, Any, Dict, Iterable, List, NamedTuple, Union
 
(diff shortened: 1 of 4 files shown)


jmtd:
I like this, but the module path stuff doesn't work how I'd expect it to. I added the following to a module descriptor `modules/run/module.yaml`:

```diff
+help:
+  add: true
+  template: my_help.md
```

where `my_help.md` I created at `modules/run`, i.e., adjacent to the `module.yaml`. This provokes an error:

    jinja2.exceptions.TemplateNotFound: 'my_help.md' not found in search path: 'target/image/modules'

What worked was to build a relative path including the modules _name_, i.e.,

```yaml
template: jboss.container.java.run/my_help.md
```

which is not what I would expect. I'd expect `my_help.md` to be resolved as a path relative to `target/image/modules/jboss.container.java.run` in this example.

I appreciate that you document exactly this behaviour ("For modules a relative path should include the module name") but, why require the module name in the relative path?

rnc:
> I like this, but the module path stuff doesn't work how I'd expect it to. I added the following to a module descriptor `modules/run/module.yaml`:

<snip>

> which is not what I would expect. I'd expect `my_help.md` to be resolved as a path relative to `target/image/modules/jboss.container.java.run` in this example.
> 
> I appreciate that you document exactly this behaviour ("For modules a relative path should include the module name") but, why require the module name in the relative path?

As you can see in https://github.com/cekit/cekit/pull/939/files#diff-af4c00a7905a314f34b84eccfef70887ec7b8f7216a7ee3bec41fbd3c80512f2R536 we prepend the path for relative path scenarios for modules. I suppose, given the module name is available in https://github.com/cekit/cekit/pull/939/files#diff-af4c00a7905a314f34b84eccfef70887ec7b8f7216a7ee3bec41fbd3c80512f2L448 (as `to_install.name`), we could append that to `"target/image/modules/"`.

I suppose this is what the initial feedback is for :-) ... if you feel like it would be better to have it without the module name I can look into that.

I should also document the potential issue that if the user has multiple modules with help, there is no way of merging so 'last' wins.

rnc:
> > I suppose this is what the initial feedback is for :-) ... if you feel like it would be better to have it without the module name I can look into that.
> 
> I think that would be a little more intuitive. Yes please! :-)

Updated already :-)

jmtd:
@migueltercero wdyt?

jmtd (approved):

PR #940, opened 2025-05-01
Clean package management cache when not using squash

### LINKED ISSUE #849: When installing packages with --no-squash the cache and lists increases the size of final images
Currently when you install packages it doesn't clean up the cache and lists after which leaves them left in the layers. They are cleaned up at the end with `apt-get clean` but without squashing this leaves the final image larger. Even with this increasing the build time a little due to no cache for `apt-get update` this will remove around 20MB from final images with a small change. 

**Additional context**
This is best practice from docker and reduces the foot guns when using the packages feature.
https://docs.docker.com/develop/develop-images/dockerfile_best-practices/

Happy to look at adding a PR to move the clean up to after each package install if that fits with the project.

rnc:
Essentially if CEKit squashing is disabled this then moves the packagement management clean next to the package install/remove/etc commands - this reduces the layers and ensures caches are not left lying around.

Fixes #849
This will also reduce the number of layers as mentioned in #593

### DIFF
--- a/cekit/builder.py
+++ b/cekit/builder.py
@@ -102,6 +102,7 @@ class Builder(Command):
             self.params.target,
             container_file,
             self.params.overrides,
+            self.params.no_squash,
         )
 
         if CONFIG.get("common", "redhat"):
--- a/cekit/builders/oci_builder.py
+++ b/cekit/builders/oci_builder.py
@@ -48,7 +48,15 @@ class OCIBuilder(Builder):
 
         cmd.append(os.path.join(self.target, "image"))
 
-        run_wrapper(cmd, False, f"Could not run {build_type} {cmd}")
+        capture: bool = False
+        if LOGGER.isEnabledFor(logging.DEBUG):
+            # Normally we'll just let output print to stdout/stderr. If verbose(debug) is enabled
+            # then we'll capture it but log it as well. This is useful for tests verifying the output.
+            capture = True
+        result = run_wrapper(cmd, capture, f"Could not run {build_type} {cmd}")
+        if result.stdout is not None:
+            LOGGER.debug(result.stdout)
+            LOGGER.debug("\n")
 
         LOGGER.info(
             f"Image built and available under following tags: {', '.join(tags)}"
--- a/cekit/generator/base.py
+++ b/cekit/generator/base.py
@@ -71,6 +71,7 @@ class Generator(object):
         target: PathType,
         container_file: str,
         overrides: List[str],
+        no_squash: bool,
     ):
         self._descriptor_path: PathType = descriptor_path
         self._overrides: List[Overrides] = []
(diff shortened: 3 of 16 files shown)


jmtd (approved):
I think this works. I tested docker and podman, --no-squash and default (squash).

I think we could go a step further: when squashing, the "_pkg manager_ clean all" could be skipped at install time and only done at the end, as well.

I.e. from (now/this PR)

```
        RUN : \
        && microdnf --setopt=install_weak_deps=0 --setopt=tsflags=nodocs install -y shadow-utils \
            && microdnf clean all && rpm -q shadow-utils \
        && :
# … now to the end
    # Clear package manager metadata
    RUN rm -rf "/var/cache/yum" "/var/lib/dnf" "/var/cache/apt" "/var/cache/dnf"
```

to

```
        RUN : \
        && microdnf --setopt=install_weak_deps=0 --setopt=tsflags=nodocs install -y shadow-utils \
            && rpm -q shadow-utils \
        && :
# … now to the end
    # Clear package manager metadata
    RUN microdnf clean all
       && rm -rf "/var/cache/yum" "/var/lib/dnf" "/var/cache/apt" "/var/cache/dnf"
```

rnc:
> I think we could go a step further: when squashing, the "_pkg manager_ clean all" could be skipped at install time and only done at the end, as well.
> 
> I.e. from (now/this PR)
> 
> ```
>         RUN : \
>         && microdnf --setopt=install_weak_deps=0 --setopt=tsflags=nodocs install -y shadow-utils \
>             && microdnf clean all && rpm -q shadow-utils \
>         && :
> # … now to the end
>     # Clear package manager metadata
>     RUN rm -rf "/var/cache/yum" "/var/lib/dnf" "/var/cache/apt" "/var/cache/dnf"
> ```
> 
> to
> 
> ```
>         RUN : \
>         && microdnf --setopt=install_weak_deps=0 --setopt=tsflags=nodocs install -y shadow-utils \
>             && rpm -q shadow-utils \
>         && :
> # … now to the end
>     # Clear package manager metadata
>     RUN microdnf clean all
>        && rm -rf "/var/cache/yum" "/var/lib/dnf" "/var/cache/apt" "/var/cache/dnf"
> ```

On the topic of clean I do wonder why we're not running `dnf clean all` (etc) i.e. why is it only `microdnf clean`? 

I am not clear why the query embeds a microdnf clean all .... as you say it might be better to move cleanup next to the lines regarding clearing the package manager metadata. That could be done for **both** squash and no_squash forms.

jmtd:
> On the topic of clean I do wonder why we're not running dnf clean all (etc) i.e. why is it only microdnf clean?

Seems like an omission.

> as you say it might be better to move cleanup next to the lines regarding clearing the package manager metadata. That could be done for both squash and no_squash forms.

For squash, the `microdnf clean` for each install costs time but doesn't save image size versus doing it once at the end. For no-squash, we'd want it done in every layer which does an install (as with now). This could be tackled in a future PR, though, and not block this one.

jmtd (approved):
LGTM!

PR #956, opened 2026-03-03
Implement Konflux builder/generator

jmtd:
This is very useful for me, but I don't know how many other users we might have.

This implements a konflux builder, very similar to the existing osbs builder. It is used to sync the generation results from cekit with another git repository. Unlike the osbs builder, It does not trigger a build, only prepares the sources.

There's a fair bit of clean up needed. Future feature: gather all packages to be installed and generate/update a `rpms.in.yaml` for hermetic builds. Possibly also generating `rpms.lock.yaml`.

 - [x] handle overrides (generator will fail if `konflux` key is missing in image yaml even if its in a specified override)
 - [x] (optionally?) resolve floating tags (e.g. `:latest`) to a `FROM img@sha256` for reproducible builds
 - [ ] (optionally?) `git push` the generated commit
 - [ ] (optionally?) import RPM packages from the base image into the `rpms.in.yaml`. Rationale: Renovate can update NVRs in our layer independently of waiting for base image respins
 - [ ] handle missing branches in output repository (create them)


cekit/generator/konflux.py
```diff
@@ -5,8 +5,22 @@
 
 from cekit.template_helper import TemplateHelper
 from cekit.generator.base import Generator
+from cekit.cekit_types import PathType
+from typing import TYPE_CHECKING, Callable, Dict, List
 
 class KonfluxGenerator(Generator):
+    def __init__(
```
jmtd: this is superfluous

